PR #13, opened 2026-03-26
feat: lazy-load scan_events via detail endpoint

rasmusfaber:
## Summary

- Make `scan_events` optional in Arrow table parsing so the bulk dataframe load works without it
- Add `getScannerDataframeDetail` API method and React Query hooks for the new `/fields` endpoint
- Wire `ScannerResultPanel` to fetch events on demand via the detail hook
- Remove dead input hooks that were replaced by the detail hook

Companion to meridianlabs-ai/inspect_scout#367 (backend).

## Test plan

- [x] Manual test: bulk scan loads without scan_events, clicking a row fetches events via `/fields` (we have deployed a prerelease of this to our infra).


apps/scout/src/app/types.ts
```diff
@@ -9,6 +9,7 @@ import {
   Event,
   JsonValue,
   ModelUsage,
+  ScannerInput,
   ScannerInputResponse,
   Transcript,
 } from "../types/api-types";
```
epatey: @rasmusfaber, I created a merge conflict here. `api-types.ts` was still mistakenly re-exporting inspect types. That wasn't my intention. I got rid of those re-exports and updated the importers to import from `inspect-common/types` directly. It should be a relatively straightforward merge resolution.

rasmusfaber: It should be resolved now.

apps/scout/src/api/api-scout-server.ts
```diff
@@ -232,35 +238,41 @@ export const apiScoutServer = (
...
+      scanner: string,
+      excludeColumns?: string[]
     ): Promise<ArrayBuffer> => {
+      const params = excludeColumns?.length
+        ? `?exclude_columns=${excludeColumns.map(encodeURIComponent).join(",")}`
+        : "";
       const result = await requestApi.fetchBytes(
         "GET",
-        `/scans/${encodeBase64Url(scansDir)}/${encodeBase64Url(scanPath)}/${encodeURIComponent(scanner)}`
+        `/scans/${encodeBase64Url(scansDir)}/${encodeBase64Url(scanPath)}/${encodeURIComponent(scanner)}${params}`
```
epatey: This is a bit of a nit and comes from me being paranoid about security and encoding wrt url's and params. In simple cases like this, it's probably fine the way it is, but here's how I might have done it.

```TypeScript
  const params = new URLSearchParams();
  if (excludeColumns?.length) {
    params.set("exclude_columns", excludeColumns.join(","));
  }
  const query = params.size ? `?${params}` : "";
  const result = await requestApi.fetchBytes(
    "GET",
    `/scans/${encodeBase64Url(scansDir)}/${encodeBase64Url(scanPath)}/${encodeURIComponent(scanner)}${query}`
  );
```

epatey (approved):
This looks good to me. It's up to you if you make that URL param change. I'll happily merge it in the AM either way - assuming you'll be ready with the Python PR as well. I'll read that one again as soon as I can.

rasmusfaber:
> This looks good to me. It's up to you if you make that URL param change. I'll happily merge it in the AM either way - assuming you'll be ready with the Python PR as well. I'll read that one again as soon as I can.

I did something crazy when trying to rebase the Python PR, that completely messed up the history. I think it is resolved, but I am not sure enough that I don't think I need a full retest. I will let it know when I am confident in the code again. (And I will clean up the URL params while I am at it. That one is just lazy code.)

apps/scout/src/api/api-scout-server.ts
rasmusfaber: Done.

rasmusfaber:
> I will let it know when I am confident in the code again.

Okay. This should be ready now. 

One thing turned up in testing, though: Very old scans (before [f69de694](https://github.com/meridianlabs-ai/inspect_scout/commit/f69de694) from 2025-11-07) do not have uuids. This breaks some assumptions made here and in the earlier lazy input loading and in a few other places.

I can just migrate the few scans we have that predate that, but I am not sure whether you want to preserve backwards compatibility that far. If you do, I guess there are a few options? Edit: for context, METR didn't have any production scans that predated this. It was just my collection of test scans that had a few.

epatey:
> One thing turned up in testing, though: Very old scans (before [f69de694](https://github.com/meridianlabs-ai/inspect_scout/commit/f69de694) from 2025-11-07) do not have uuids. This breaks some assumptions made here and in the earlier lazy input loading and in a few other places.

This is something we're going to have to deal with. I'm guessing that type hints and static analysis didn't keep us honest because we're talking about pulling these out of Parquet/dataframe columns. Is that right?

epatey:
I've been reading the code a bit more. Although it's fragile because of going through arrow, we're not actually losing type safety. The issue you're suggesting is that for old scans without a UUID, we have no way to go back later and retrieve `input` and `scan_events`. We have no key to do the lookup with. Is that it?

PR #14, opened 2026-03-27
fix: disable polling and topic invalidation for completed scans

rasmusfaber:
## Summary

When viewing a completed scan, the viewer continues making network requests in the background to check for updates.

This fixes the issue in two ways:

1. Just drop the refetching when viewing a completed scan
2. Drop `scans-inv` from the query keys for completed scans (`useScan` bootstraps from the query cache to detect completion; `useScanDataframe` takes `isComplete` from the parent hook)

There might still be a remaining issue with SSE reconnection triggering query invalidations. Claude claims that:

> Every SSE reconnect (or every 10s poll response when SSE is disabled behind a reverse proxy via window.__SCOUT_DISABLE_SSE__) sends the full topic state as its first message. The frontend's prevVersionsRef starts as undefined, so each reconnect triggers undefined !== "<timestamp>" → invalidates all queries keyed with "scans-inv". This PR prevents that from affecting completed scan data (those queries no longer carry "scans-inv"), but active scans and the scan list still get unnecessary refetches on every reconnect. A proper fix would seed prevVersionsRef from the initial SSE message without triggering invalidation.

But I am not addressing this here.

Companion PR: https://github.com/meridianlabs-ai/inspect_scout/pull/370


epatey:
Before changing the cache behavior here, I  need to better understand the concrete scenario this PR is trying to fix. Is the concern that the scans endpoint is expensive, or that completed scan pages are causing meaningful
cache / network churn?

#### `ScanPanel` and `useScans`

Separately,  I do not think `ScanPanel` should depend on `useScans`.

Right now it is only used for `loading: scansLoading`. This looks like leftover coupling from the pre-React Query implementation of the viewer. At this point, the loading state from `useScan` should be sufficient.

Letting the caller control polling behavior exposes custom hook implementation details to compensate for a page-level design smell.

I think the cleaner fix is to remove `useScans` from `ScanPanel` entirely and let the scan detail view depend on `useSelectedScan` / scan-specific queries only.

If scan polling policy needs to change, that should be encapsulated inside `useScan`.

#### Be Careful About Dropping `scans-inv`

The `scans` invalidation topic is fired when the collection of scans changes semantically, including:
- a scan being deleted
- a scan reaching the normal terminal path and updating the last-scan signal

This PR changes the contract by removing completed `useScan` and `useScanDataframe` queries from the `scans-inv` invalidation set, while other scan-related queries still participate.

At minimum, that seems risky for deletion. On `main`, a completed scan view would still be invalidated if the scan were deleted; with this PR, completed queries that have dropped `scans-inv` can remain pinned to stale cache entries.

So my high-level feedback is:
1. Please clarify the concrete scenario we are trying to fix.
2. Remove `useScans` from `ScanPanel` rather than exposing its polling controls.
3. If scan polling policy needs to change, encapsulate that inside `useScan`.
4. Be careful about weakening `scans-inv` semantics for completed scan queries.

rasmusfaber:
Closing in favor of the better fix in #46

PR #22, opened 2026-04-01
Improve `ScanResultSummary` filtering 

brandly:
@dragonstyle and I noticed some surprising behavior when searching through scan results. Each of these results starts with "The agent demonstrates" but searching for "agents" produces no results.

| Full list | Searching for "agents" |
| ---- | ---- |
| Screenshot 2026-04-01 at 12 28 45 PM 1 | Screenshot 2026-04-01 at 12 28 59 PM |

With these changes, we'll look for more exhaustively through `ScanResultSummary` for text that matches the search query.

Screenshot 2026-04-01 at 12 31 33 PM

I added tests for the new `stringifyValue` and some other existing functions


dragonstyle (approved):
👍

PR #44, opened 2026-04-07
Migrate Inspect to Vitest

brandly:
I started to go down a rabbit hole of mocking out modules to add a Jest test, but the rest of the repo already uses Vitest! Seems like we should migrate instead.

Vitest provides a nice guide and makes it easy: https://vitest.dev/guide/migration.md


dragonstyle:
This is great - I ran into an issue with testing in one of my branches and converting to vitest was my solution! That wouldn't have landed for a day or two so this is better! 🔥

PR #46, opened 2026-04-07
Remove dead useScans hook; ScanPanel uses useSelectedScan loading state instead

epatey:


dragonstyle:
Question - was this need for 'reloading' or some scenario where the page got loaded without hitting previous routes?

epatey:
> Question - was this need for 'reloading' or some scenario where the page got loaded without hitting previous routes?

@dragonstyle, I think there were a couple of factors that led us here.

First, when I introduced react-query, I wanted to be be very careful to honor the _loading_ logic that had historically been in Zustand.

Second, this was originally written before we had the paginated infinite query. In retrospect, `useScans` should have been deleted when we switched. 

Even if there were some side-effect of `useScans` that `ScanPanel` depended upon (which I'll bottom out on), it would be a smell. `useSelectedScan` should be sufficient. If it has dependencies that it needs to make itself work, it should encapsulate them.

PR #49, opened 2026-04-08
Loudly warn when `pnpm watch` is run outside context of a submodule mount.

epatey:
## Summary
- Add red/bold warning when `pnpm watch` is run without a submodule mount (dist copy has no effect in that context)
- Centralize the Vite plugin in `tooling/python-repo/index.js` alongside existing submodule detection helpers


dragonstyle:
🙏

PR #52, opened 2026-04-10
feat: stream deduplicated message pools during in-progress evals

rasmusfaber:
## Summary

- Add `message_pool` and `call_pool` fields to `SampleData` and pool cursor parameters to all API layers (view-server, vscode, client-api)
- Accumulate pool entries in `samplePolling` across polling iterations and resolve condensed `ModelEvent` input/call refs client-side during live streaming
- Fix `existingIndex === 0` falsy check bug in event replacement

Companion PR to https://github.com/UKGovernmentBEIS/inspect_ai/pull/3375 which adds the Python-side buffer pool dedup and server endpoints.

## Test plan

Tested with https://github.com/UKGovernmentBEIS/inspect_ai/pull/3375

### DIFF
--- a/apps/inspect/src/client/api/types.ts
+++ b/apps/inspect/src/client/api/types.ts
@@ -2,6 +2,7 @@ import type {
   ApprovalEvent,
   AttachmentData,
   BranchEvent,
+  CallPoolData,
   CompactionEvent,
   ErrorEvent,
   EvalError,
@@ -20,6 +21,7 @@ import type {
   LogHandle,
   LogInfo,
   LogUpdate,
+  MessagePoolData,
   ModelEvent,
   SampleInitEvent,
   SampleLimitEvent,
@@ -39,21 +41,7 @@ import {
   EvalSampleTarget,
 } from "../../@types/extraInspect";
 
-export interface MessagePoolData {
-  id: number;
-  sample_id: string;
-  epoch: number;
-  msg_id: string;
-  data: string;
-}
-
-export interface CallPoolData {
-  id: number;
-  sample_id: string;
-  epoch: number;
-  hash: string;
-  data: string;
-}
+export type { CallPoolData, MessagePoolData };
 
 // Hand-coded — references the local EventData with typed event union
 export interface SampleData {
--- a/apps/inspect/src/client/api/types.ts
+++ b/apps/inspect/src/client/api/types.ts
@@ -39,10 +39,28 @@ import {
   EvalSampleTarget,
 } from "../../@types/extraInspect";
 
+export interface MessagePoolData {
+  id: number;
+  sample_id: string;
+  epoch: number;
+  msg_id: string;
+  data: string;
+}
+
+export interface CallPoolData {
+  id: number;
+  sample_id: string;
+  epoch: number;
+  hash: string;
+  data: string;
+}
+
 // Hand-coded — references the local EventData with typed event union
 export interface SampleData {
   events: EventData[];
   attachments: AttachmentData[];
+  message_pool: MessagePoolData[];
+  call_pool: CallPoolData[];
 }
 
 export type ProgressCallback = (
@@ -188,7 +206,9 @@ export interface LogViewAPI {
     id: string | number,
     epoch: number,
     last_event?: number,
-    last_attachment?: number
+    last_attachment?: number,
+    last_message_pool?: number,
+    last_call_pool?: number
   ) => Promise<SampleDataResponse | undefined>;
 }
 
@@ -233,7 +253,9 @@ export interface ClientAPI {
     id: string | number,
     epoch: number,
     last_event?: number,
-    last_attachment?: number
+    last_attachment?: number,
+    last_message_pool?: number,
+    last_call_pool?: number
   ) => Promise<SampleDataResponse | undefined>;
 
   // Events
--- a/apps/inspect/src/state/samplePolling.ts
+++ b/apps/inspect/src/state/samplePolling.ts
@@ -298,10 +298,7 @@ function processMessagePool(
   sampleData: SampleData,
   pollingState: PollingState
 ) {
-  if (!sampleData.message_pool?.length) return;
-  log.debug(
-    `Processing ${sampleData.message_pool.length} message pool entries`
-  );
+  if (!sampleData.message_pool.length) return;
   for (const entry of sampleData.message_pool) {
     pollingState.messagePool.push(JSON.parse(entry.data) as ChatMessage);
     pollingState.messagePoolId = Math.max(pollingState.messagePoolId, entry.id);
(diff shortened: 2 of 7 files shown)


apps/inspect/src/client/api/types.ts
```diff
@@ -39,10 +39,28 @@ import {
...
+}
+
+export interface CallPoolData {
+  id: number;
+  sample_id: string;
+  epoch: number;
+  hash: string;
+  data: string;
+}
+
```
epatey: You shouldn't need to manually recreate these python types. If they're referenced by the FastAPI server, they should be included in the generated OpenAPI schema as well as the generated TypeScript types.

You'd need to add code to `packages/inspect-common/src/types/index.ts` pluck them out of `generated.ts` and re-export them instead.

apps/inspect/src/state/samplePolling.ts
```diff
@@ -269,6 +294,29 @@ function processAttachments(
...
 }
 
+function processMessagePool(
+  sampleData: SampleData,
+  pollingState: PollingState
+) {
+  if (!sampleData.message_pool?.length) return;
+  log.debug(
+    `Processing ${sampleData.message_pool.length} message pool entries`
+  );
```
epatey: This probably wants to get deleted when you're done.

apps/inspect/src/state/samplePolling.ts
```diff
@@ -269,6 +294,29 @@ function processAttachments(
   });
 }
 
+function processMessagePool(
+  sampleData: SampleData,
+  pollingState: PollingState
+) {
+  if (!sampleData.message_pool?.length) return;
```
epatey: The `?` shouldn't be needed. We're still using lenient lint rules in `apps/inspect`. This would have been flagged in other packages.

apps/inspect/src/state/samplePolling.ts
```diff
@@ -269,6 +294,29 @@ function processAttachments(
...
+  );
+  for (const entry of sampleData.message_pool) {
+    pollingState.messagePool.push(JSON.parse(entry.data) as ChatMessage);
+    pollingState.messagePoolId = Math.max(pollingState.messagePoolId, entry.id);
+  }
+}
+
+function processCallPool(sampleData: SampleData, pollingState: PollingState) {
+  if (!sampleData.call_pool?.length) return;
+  log.debug(`Processing ${sampleData.call_pool.length} call pool entries`);
```
epatey: Delete this when you're done.

apps/inspect/src/state/samplePolling.ts
```diff
@@ -286,7 +334,7 @@ function processEvents(
     const existingIndex = pollingState.eventMapping[eventData.event_id];
 
     // Resolve attachments within this event
-    const resolvedEvent = resolveAttachments<Event>(
+    let resolvedEvent = resolveAttachments<Event>(
```
epatey: I really hate `let`'s. I'd prefer it if you could either:
- combine this all into a single statement assigning into a `const resolvedElement` 
or
- have 3 `const` variables that build upon each other.

apps/inspect/src/state/samplePolling.ts
```diff
@@ -324,6 +382,51 @@ function processEvents(
...
 
+function expandRefs<T>(refs: [number, number][], pool: T[]): T[] {
+  return refs.flatMap(([start, end]) => pool.slice(start, end));
+}
+
+function resolvePoolRefs(event: Event, pollingState: PollingState): Event {
+  const ev = event as ModelEvent;
+  if (ev.event !== "model") return event;
+
+  let resolved = ev;
```
epatey: Prefer type narrowing over casting.
```suggestion
  if (event.event !== "model") return event;
  let resolved = event;
```

Like above, I'd also prefer `const` over `let`. In this case, it might be worth it to have helper functions that enable a `const` construction.

apps/inspect/src/state/samplePolling.ts
```diff
@@ -324,6 +382,51 @@ function processEvents(
...
+
+  if (
+    Array.isArray(resolved.input_refs) &&
+    pollingState.messagePool.length > 0
+  ) {
+    resolved = {
+      ...resolved,
+      input: expandRefs(
+        resolved.input_refs as [number, number][],
```
epatey: I don't think this cast is required.

apps/inspect/src/state/samplePolling.ts
```diff
@@ -324,6 +382,51 @@ function processEvents(
...
+        pollingState.messagePool
+      ) as ModelEvent["input"],
```
epatey:
```suggestion
      ) satisfies ModelEvent["input"],
```
Casting is lame.

apps/inspect/src/state/samplePolling.ts
```diff
@@ -324,6 +382,51 @@ function processEvents(
...
+      input_refs: null,
+    };
+  }
+
+  if (resolved.call && Array.isArray(resolved.call.call_refs)) {
+    const msgKey = (resolved.call.call_key as string) || "messages";
+    const request = { ...resolved.call.request };
+    request[msgKey] = expandRefs(
+      resolved.call.call_refs as [number, number][],
```
epatey: same as above - you can get rid of the cast

apps/inspect/src/state/samplePolling.ts
rasmusfaber: It is not just for linting. message_pool can legitimately be undefined if this runs against an older server.

But perhaps that is not an issue anymore after the VSCode plugin changes you told me about? I have removed it here and in `processCallPool()`, but please double check this.

apps/inspect/src/state/samplePolling.ts
rasmusfaber: Done

apps/inspect/src/client/api/types.ts
rasmusfaber: Ah, yes. Fixed.

apps/inspect/src/state/samplePolling.ts
rasmusfaber: It is gone.

apps/inspect/src/state/samplePolling.ts
rasmusfaber: Gone.

apps/inspect/src/state/samplePolling.ts
rasmusfaber: Done.

apps/inspect/src/state/samplePolling.ts
rasmusfaber: Nice! I haven't had a chance to use that before.

apps/inspect/src/state/samplePolling.ts
rasmusfaber: Gone.

apps/inspect/src/state/samplePolling.ts
epatey: Interesting. That complicates things a little bit. I have a principle about trusting the type system. I avoid add runtime defensiveness for things the compiler already guarantees. So in a case like this, we could/should repair the data upstream at the point where I assume we currently cast it.

apps/inspect/src/state/samplePolling.ts
```diff
@@ -254,8 +275,12 @@ export function createSamplePolling(
 const resetPollingState = (state: PollingState) => {
   state.eventId = kNoId;
   state.attachmentId = kNoId;
+  state.messagePoolId = kNoId;
+  state.callPoolId = kNoId;
```
epatey, line 279: I know this isn't new with you PR, but I need to go on record that `kNoId` is a smell in my opinion. It's doing what `| undefined` is for. Hmmm.

PR #78, opened 2026-04-18
fix: break infinite re-render loop in bulk collapse/expand

rasmusfaber:
## Summary

When pressing the collapse button in the transcript tab, you entered an infinite loop resulting in React error #185 (maximum update depth exceeded).

- Root cause: `onSetTranscriptCollapsed` callback was included in the `useEffect` dependency array, but calling it updates `collapsedEvents` in the Zustand store, which causes `collapseState` to recalculate via `useMemo`, creating a new callback reference, which re-triggers the effect — infinite loop
- Fix: use a `useRef` to hold the `onSetTranscriptCollapsed` callback, breaking the circular dependency

## Test plan

- [x] Verified locally that clicking Collapse no longer causes React error #185
- [x] Verified that Collapse collapses all collapsible sections (step, span_begin, tool, subtask)
- [x] Verified that Expand restores sections to their default collapsed state
- [x] Run existing tests

### DIFF
--- a/packages/inspect-components/src/transcript/TranscriptLayout.tsx
+++ b/packages/inspect-components/src/transcript/TranscriptLayout.tsx
@@ -22,6 +22,7 @@ import {
   useCallback,
   useEffect,
   useMemo,
+  useRef,
   useState,
 } from "react";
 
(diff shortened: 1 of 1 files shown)


dragonstyle:
Thanks for this - I went after the root cause + fixed a few other issues in a PR that superseded this one.

rasmusfaber:
> Thanks for this - I went after the root cause + fixed a few other issues in a PR that superseded this one.

Even better. Thanks!

PR #80, opened 2026-04-20
fix: stabilize collapse callbacks to prevent infinite re-render loop

dragonstyle:
## Summary

- The `collapseState` useMemo bundled callbacks with data fields, all depending on `collapsedEvents`. When the bulk-collapse effect called `onSetTranscriptCollapsed`, it updated `collapsedEvents`, which recreated the memo — giving the callback a new reference that re-triggered the effect (React error #185)
- Extract callbacks into `useCallback` with stable Zustand/immer action deps so their identity never changes when `collapsedEvents` updates
- Supersedes the `useRef` workaround in #78 by fixing the root cause: callbacks were needlessly coupled to data changes

## Test plan

- [x] `pnpm check` and `pnpm build` pass
- [x] Click Collapse — all sections collapse, no React error #185
- [x] Click Expand — sections restore to defaults
- [x] Individual collapse/expand toggles still work after bulk operation
- [x] Switching samples resets collapse state correctly

### DIFF
--- a/apps/inspect/src/app/samples/transcript/TranscriptPanel.tsx
+++ b/apps/inspect/src/app/samples/transcript/TranscriptPanel.tsx
@@ -115,21 +115,44 @@ export const TranscriptPanel: FC<TranscriptPanelProps> = memo((props) => {
     (state) => state.sampleActions.collapseEvent
   );
 
+  const onCollapseTranscript = useCallback(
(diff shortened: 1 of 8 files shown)


dragonstyle:
This is an alternative way to address the issue resolved in #78 without the need for a ref guard.

PR #93, opened 2026-04-21
fix: scope filtering of retried logs to the folder

ransomr:
The viewer does not show retried logs be default. The deduplication is done at the log_dir level. The issue is that when you have a log_dir with two subdirectories, each of which has a copy of the same log file. One of the subdirs will show up as empty because the only log in that directory is shown as being retried.

This changes the retried check to scope to the directory, so that the copied log will no longer show as retried.


apps/inspect/src/state/hooks.ts
```diff
@@ -545,60 +545,80 @@ const simplifiedStatusForDeduplication = (status: EvalLogStatus | undefined) =>
...
+  // If same priority, take the latest one
+  const bestByName: Record<string, LogHandleWithretried> = {};
+  for (const items of Object.values(logsByGroup)) {
+    items.sort((a, b) => {
+      const as = simplifiedStatusForDeduplication(logPreviews[a.name]?.status);
+      const bs = simplifiedStatusForDeduplication(logPreviews[b.name]?.status);
+      const am = a.mtime ?? 0;
+      const bm = b.mtime ?? 0;
+
+      if (as === bs) return bm - am; // newest on top
```
dragonstyle: I just merged a Rasmus PR that switches this to tie breaking on name which has ISO timestamp.

https://github.com/meridianlabs-ai/ts-mono/pull/96

Can you incorporate that here?

dragonstyle (changes requested):

apps/inspect/src/state/hooks.ts
ransomr: merged

dragonstyle (approved):

PR #99, opened 2026-04-22
Improve Markdown Rendering Safety

dragonstyle:
1) Protect specifically against shell style interpolation being interpreted as math.

2) Try to balance backticks (conservatively).

3) Extract render pipeline to shared location, add tests.


epatey:
@dragonstyle  I've tried to establish a pattern of locating test files right next to the source files rather than in a separate directory. I happy to chat about the benefits. It seems like I could add a bullet in our CLAUDE.md to suggest that preference.

dragonstyle:
> I've tried to establish a pattern of locating test files right next to the source files rather than in a separate directory.

doh! We've talked about this, I just missed it! Will fix.

brandly:
This was driven by trying to render markdown in search results. Search results show truncated text, so we can end up with unbalanced backticks.

However after more thinking, we've concluded that we probably just shouldn't render rich markdown in search results. Some situations are definitely better with markdown rendering, but it's fraught with peril. Too many edge cases can blow up the parser and lead to really ugly rendering.

PR #135, opened 2026-04-26
CSS module types

brandly:
I was surprised that accessing `styles.notDefined` didn't produce a type error. The root of the issue is the `@types/css-modules` package defines `styles` as `{ [key: string]: string }`. Claude researched and implemented these changes.

We're now generating a `.d.ts` file for each `.module.css`, and we `.gitignore` these files since they're derived from the `.module.css` files. The fact that we have to write our own integration between `typed-css-modules` and Vite's dev server make me somewhat skeptical that this is the right approach, but with it, we get live feedback as we edit files and autocomplete as we begin to access a classname. The developer experience is definitely better, so it's mostly a question of whether or not this is the best approach.

https://github.com/user-attachments/assets/22a28ba8-5f99-4af6-b3fe-9ece98fd92e2

Introducing this produced quite a few type errors, and Claude cleaned them up with the goal of avoiding any styling impacts.


dragonstyle:
As a follow up - I did a little more reading about some of those libraries that we explored today and I think css modules remains a fine direction for now. We'll be able to get to a stable css var based theme (but that is typed/lint checked). 

1) Claude suggested that `tcm --watch` would be a replacement for that tooling `js` that you were skeptical about...

2) I think we should consider https://github.com/mrmckeb/typescript-plugin-css-modules as a way to not require the dev server to get intellisense for styles...

Claude nits:

- Four identical src/types/css.d.ts shims could live in one shared place (e.g. @tsmono/tsconfig exports or a @tsmono/types package) and be referenced from each tsconfig — small de-dup.
- tooling/css-modules-dts/ isn't declared as a workspace package; it's imported by relative path. If you keep it, give it a package.json and reference it as @tsmono/css-modules-dts so it's discoverable and the relative path doesn't quietly rot.
- turbo.json: build doesn't depend on css-types. Fine if build is vite build only, but if any package's build invokes tsc (e.g. for dist/ types), the dep is missing.

PR #138, opened 2026-04-26
Fix message ordering in running-sample Messages tab

rasmusfaber:
## Summary

While a sample is running, the Messages tab built its list by walking model events in stream order with a `Map<id, message>`. `Map.set` on an existing key keeps the original insertion position, so tool results and follow-up user turns from later events were appended at the end instead of slotting between the assistants that produced them. The bug self-healed once the sample completed and the server-provided `sample.messages` was used instead.

This PR walks model events in order and reconstructs the conversation incrementally. Known messages in each event's `input` anchor a cursor to their position; new messages splice in at `cursor + 1`. Outputs append. The cursor defaults to the end of the list so a fresh first message in an event (e.g. a compaction summary that the post-compaction event introduces ahead of any prior anchor) lands after the existing prefix rather than at the front.

This handles three scenarios:

1. **Original bug — tool results clumped at the end.** When tool results first appear in a later event's `input` interleaved among earlier assistants, they splice immediately after the assistant whose call they answer, not at the end.
2. **Compaction.** Post-compaction events have a shortened `input` that omits earlier messages. Those earlier messages stay visible (they're already in the running list), and the new compaction summary slots in after them.
3. **Intermediate outputs not folded into a later input.** An assistant produced by an event whose output never appears in a subsequent event's `input` lands between its event's last input anchor and the next event's new messages.

## Test plan

- [X] Manual: load a running sample with tool calls and verify Messages tab interleaves assistants/tools/users correctly; verify a compacting sample preserves pre-compaction history.

### DIFF
--- a/apps/inspect/src/app/samples/messagesFromEvents.ts
+++ b/apps/inspect/src/app/samples/messagesFromEvents.ts
@@ -1,41 +1,58 @@
 import type {
   ChatMessage,
-  ChatMessageAssistant,
-  ChatMessageSystem,
-  ChatMessageTool,
-  ChatMessageUser,
   Event,
+  ModelEvent,
 } from "@tsmono/inspect-common/types";
 
+const isSuccessfulModelEvent = (e: Event): e is ModelEvent =>
+  e.event === "model" && !e.error;
+
 export const messagesFromEvents = (runningEvents: Event[]): ChatMessage[] => {
-  const messages: Map<
-    string,
-    ChatMessageSystem | ChatMessageUser | ChatMessageAssistant | ChatMessageTool
-  > = new Map();
+  const modelEvents = runningEvents.filter(isSuccessfulModelEvent);
 
-  runningEvents
-    .filter((e) => e.event === "model")
-    .filter((e) => !e.error)
-    .forEach((e) => {
-      for (const m of e.input) {
-        const inputMessage = m as
-          | ChatMessageSystem
-          | ChatMessageUser
-          | ChatMessageAssistant
-          | ChatMessageTool;
-        if (inputMessage.id && !messages.has(inputMessage.id)) {
-          messages.set(inputMessage.id, inputMessage);
-        }
-      }
-      const outputMessage = e.output.choices[0].message;
-      if (outputMessage.id) {
-        messages.set(outputMessage.id, outputMessage);
-      }
-    });
+  // Build the conversation list by merging each model event's `input`
+  // (the model's view at that call) in order. Per event: reset a cursor
+  // to the end of the list, then walk its input. Known ids re-anchor
+  // the cursor to their position without re-inserting; new ids splice
+  // at cursor + 1. The output appends at the end of the list (not at
+  // cursor + 1) and dedupes by id.
+  //
+  // The end-of-list cursor default lets an event whose input begins
+  // with unseen messages (e.g. a compaction summary introduced ahead
+  // of any prior anchor) extend the prefix rather than prepend. The
+  // output-at-end rule then keeps the post-compaction assistant after
+  // the summary, not between it and earlier history.
+  const result: ChatMessage[] = [];
+  const positions = new Map<string, number>();
 
-  if (messages.size > 0) {
-    return messages.values().toArray();
-  } else {
-    return [];
+  const insert = (index: number, m: ChatMessage) => {
+    result.splice(index, 0, m);
+    for (const [id, pos] of positions) {
+      if (pos >= index) positions.set(id, pos + 1);
+    }
+    if (m.id) positions.set(m.id, index);
+  };
+
+  for (const e of modelEvents) {
+    let cursor = result.length - 1;
+    const seenInEvent = new Set<string>();
+    for (const m of e.input) {
+      if (!m.id || seenInEvent.has(m.id)) continue;
+      seenInEvent.add(m.id);
+      const known = positions.get(m.id);
+      if (known !== undefined) {
+        cursor = known;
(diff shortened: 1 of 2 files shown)


apps/inspect/src/app/samples/messagesFromEvents.ts
```diff
@@ -1,41 +1,58 @@
...
+    const seenInEvent = new Set<string>();
+    for (const m of e.input) {
+      if (!m.id || seenInEvent.has(m.id)) continue;
+      seenInEvent.add(m.id);
+      const known = positions.get(m.id);
+      if (known !== undefined) {
+        cursor = known;
+      } else {
+        cursor += 1;
+        insert(cursor, m);
```
dragonstyle, line 47: Do you have any concern about iterating the positions while iterating the model events (e.g. at the end of long running evals of 1k turns)?

rasmusfaber: Yes. Quite a few actually. I did some measurements, and for 1k turns this takes on the order of 30ms per poll. If you scale up to 50k turns it starts being costly with around 65s per poll.

We _can_ make this linear. Something like [this](https://github.com/METR/ts-mono/compare/fix/running-sample-message-order...39422c7) is linear time, but it is quite a lot of complexity to add. But I would be happy to polish that off instead if you prefer. We are probably going to need it sooner rather than later as runs scale up. What do you think?

dragonstyle: Claude's suggestion here is that if we use a doubly linked link to build the list of messages, we can make this very efficient and just way head->tail to build the array once the message positions are computed:

sketch:
```typescript
import type { ChatMessage, Event, ModelEvent } from "@tsmono/inspect-common/types";

type Node = { msg: ChatMessage; prev: Node | null; next: Node | null };

const isSuccessfulModelEvent = (e: Event): e is ModelEvent =>
  e.event === "model" && !e.error;

export const messagesFromEvents = (runningEvents: Event[]): ChatMessage[] => {
  const modelEvents = runningEvents.filter(isSuccessfulModelEvent);

  let head: Node | null = null;
  let tail: Node | null = null;
  const byId = new Map<string, Node>();

  // Insert `msg` immediately after node `after`. `after === null` means
  // "insert at head" (only used when the list is empty).
  const insertAfter = (after: Node | null, msg: ChatMessage): Node => {
    const node: Node = {
      msg,
      prev: after,
      next: after ? after.next : head,
    };
    if (after) {
      if (after.next) after.next.prev = node;
      else tail = node;
      after.next = node;
    } else {
      if (head) head.prev = node;
      else tail = node;
      head = node;
    }
    if (msg.id) byId.set(msg.id, node);
    return node;
  };

  const append = (msg: ChatMessage): Node => {
    const node: Node = { msg, prev: tail, next: null };
    if (tail) tail.next = node;
    else head = node;
    tail = node;
    if (msg.id) byId.set(msg.id, node);
    return node;
  };

  for (const e of modelEvents) {
    let cursor: Node | null = tail; // end-of-list default
    const seenInEvent = new Set<string>();
    for (const m of e.input) {
      if (!m.id || seenInEvent.has(m.id)) continue;
      seenInEvent.add(m.id);
      const known = byId.get(m.id);
      cursor = known ?? insertAfter(cursor, m);
    }
    const out = e.output.choices[0]?.message;
    if (out?.id && !byId.has(out.id)) append(out);
  }

  const result: ChatMessage[] = [];
  for (let n = head; n; n = n.next) result.push(n.msg);
  return result;
};

```

rasmusfaber: Yes. My Claude suggested a linked list as well (it is pretty much [this](https://github.com/METR/ts-mono/compare/fix/running-sample-message-order...8f78409)), but measurements don't really bear it out. Real data doesn't really have many splices, so it is not there the real cost is. The linked list is actually slower onThe many of the real examples I measured.

The one I link above does just the array+Map+reindex, but makes sure that we cache the results, so we don't perform extra work when hot polling.

Yet another approach would be to embrace the fact that almost all model events just is a simple append of one or more messages. We can make the 99.9% case super fast, and then do some simple checks to detect the 0.1% case. Claude seems to think that the following list is exhaustive for shapes Inspect emits:

> Choose full walk when any of these fail:
>   1. A CompactionEvent was emitted between the previous and current model event (catches
>   edit-style compaction that keeps length+last-id stable but modifies a middle message).
>   2. e.input.length < prev.length (catches summary/trim compaction).
>   3. e.input[prev.length - 1].id !== prev.lastInputId (catches the original "21 assistants then
>    late tool fold" pattern; tools at position prev.length-1 don't match the assistant id we
>   expected).
>   4. e.input[prev.length].id !== prev.outputId (catches the intermediate-output case from
>   scenario C: when prev's output isn't folded into the next call's input, the chain has
>   effectively forked).

[This](https://github.com/meridianlabs-ai/ts-mono/compare/main...7b25dbc) is an implementation. It is much, much faster than the alternative (600x at 5k turns), but you probably could trick it if you carefully crafted model events with replaced messages in the middle.

dragonstyle:
Whoops, I'm sorry - I was intending to squash merge another PR and merged the wrong PR. I'm going to revert this PR.

Your original suggestion (reducing work across polls) seems like the best path. Could you revive this PR with that?

rasmusfaber:
> Your original suggestion (reducing work across polls) seems like the best path. Could you revive this PR with that?

Sure. I will make a new PR with that implementation. Edit: #169

PR #155, opened 2026-04-28
Transcript search panel

brandly:
Clicking "Search" on a transcript's toolbar will open a form that allows users to run LLM/grep searches over the current transcript. Underneath the hood, this uses the existing LLM/grep scanners.

Past searches and their results are stored in the KVStore. This allows us to show a list of recent searches and allows us to quickly show past search results, which is particularly useful in the LLM case where we can show you cached tokens instead of paying for more.

We store a list of models used in user settings and provide previously-used models in an autocomplete input dropdown, allowing you to quickly switch between the models you like.

Screenshot 2026-04-29 at 10 25 29 AM

### Implementation notes

Grep search results include truncated text to highlight the match and show some its surrounding context. This caused a lot of issues with `MarkdownDivWithReferences`, as the truncation often leads to broken markdown. `MarkdownDivWithReferences` now has a `renderer` prop that has replaced its `omitMath` and `omitMedia` props. Passing `renderer="fragment"` will skip the more complicated rendering pipeline and will only give us basic things like linebreaks and emphasis.

~~Previously opening the validation sidebar would but `?validation=1` in the URL. The validation and search sidebars are now mutually exclusive, so instead we'll use `?sidebar=validation` and `?sidebar=search`.~~ We still support `?validation=1` and now also support `?sidebar=1`. You can have both open if you want!


packages/react/src/components/markdownRendering.ts
```diff
@@ -242,3 +247,64 @@ export function unescapeCodeHtmlEntities(str: string): string {
...
+  // For `code` tags, reverse the escaping if we can
+  const withCode = unescapeCodeHtmlEntities(unescaped);
+
+  // For `sup` tags, reverse the escaping if we can
+  const withSup = unescapeSupHtmlEntities(withCode);
+
+  return withSup;
+};
+
+const renderSimpleMarkdown = (markdown: string): string => {
```
dragonstyle: I don't think this is used anywhere...

packages/react/src/components/markdownRendering.ts
```diff
@@ -9,8 +9,9 @@ import markdownitMathjax3 from "markdown-it-mathjax3";
 // Module-level cache for lazy-initialized markdown-it instances
 const mdInstanceCache: Record<string, MarkdownIt> = {};
 
-const getOptionsKey = (omitMedia?: boolean, omitMath?: boolean): string =>
-  `${omitMedia ? "1" : "0"}:${omitMath ? "1" : "0"}`;
+export type MarkdownRenderer = "full" | "compact" | "simple";
```
dragonstyle: how about `full` | `textOnly` | `fragment` (or `minimal`)?

apps/scout/src/app/timeline/components/TimelineEventsView.tsx
```diff
@@ -60,6 +61,8 @@ interface TimelineEventsViewProps {
   getEventUrl?: (eventId: string) => string | undefined;
   /** Whether deep-link copy buttons are enabled. */
   linkingEnabled?: boolean;
+  /** Extra context fields merged into every EventNodeContext entry. */
+  eventNodeContext?: Partial<EventNodeContext>;
```
dragonstyle: Sort of thinking a narrower type might be better here (e.g. something like messageLabels)? Can construct the eventnodecontext internally...

apps/scout/src/app/transcript/SearchPanel.tsx
```diff
@@ -0,0 +1,682 @@
...
+                onClick={handleNewSearch}
+                title="New search"
+                aria-label="New search"
+              >
+                <i className={ApplicationIcons.add} aria-hidden="true" />
+              </button>
+            </div>
+          </div>
+          <div className={styles.inputShell}>
+            {/* TODO: grow up to 10 lines tall and then scroll */}
```
dragonstyle: If a simple tweak, not a bad idea for longer prompts.

dragonstyle (commented):
Taking it for a spin now!

dragonstyle:
When I first used the search panel, I have no model set (or in my env). The placeholder text just says 'filter'.

Screenshot 2026-04-30 at 8 42 41 PM

The first run experience is going to be bumpy / annoying I wonder if we should do something to try to bootstrap up a model as a fallback (e.g. look at the model in the transcript)?

dragonstyle:
Topic - we should chat about progress / super giant trajectories. I just search a 371m token trajectory and it was slow enough that progress was pretty bad. We should chat b/c may just not be realistic to bite off right now. 

When I did finally get results, I think there were none?:

Screenshot 2026-04-30 at 8 49 57 PM

dragonstyle:
Looking great overall! Some feedback from taking it for a spin...

## Search panel — consolidated review feedback

Grouping the loose comments into a checklist so they can be tracked. Each item references the original screenshot.

### UX / first-run
- [x] **Placeholder reads "filter" when no model is set / in env.** First-run experience is going to be bumpy — wonder if we should bootstrap a fallback model (e.g. look at the model in the transcript)?

EDIT: I tried this out, but at least for Claude Code transcripts, the `model` on the transcript is just like `claude-opus-4-7` instead of `anthropic/claude-opus-4-7`.

  Screenshot 2026-04-30 at 8 42 41 PM

### Visual polish
- [x] **Drop the dotted lines and add a search icon to the left of the "searching..." text.**

  Screenshot 2026-04-30 at 8 45 58 PM

- [x] **Loading/progress bar escaping the sidebar** — appears over content in the left column. (My bug, was relying on it being at the viewport edge.)

  Screenshot 2026-04-30 at 8 47 28 PM

- [x] **Dropdown should be closer / more attached to the select box.**

  Screenshot 2026-04-30 at 9 01 07 PM

- [x] **VS Code dark mode** — "searching" text and "results" label aren't the correct color. Icons for **previous searches** and **new search** are also the wrong color.

### Behavior
- [x] **Allocate space for message labels even when a message has no label** — so labeled and unlabeled messages line up vertically in transcript view. (Also affects other results views.)

  Screenshot 2026-04-30 at 8 59 59 PM

- [x] **Message labels not showing in transcript when grep scanner has results.**

- [x] **Global arrow keys (navigation) still capture left/right when text input has focus.** Should be disabled / skipped when focus is in an input — same fix pattern as the validation sidebar.

- [x] **Clicking cite labels like `[E4]` in VS Code breaks the viewer.** Probably related to needing to not actually route through the URL.

- [x] I think we should leave the search prompt populated as the user navigates using the next / previous navigation (probably also the results if available).

- [x] For llm searches at least, it's sort of weird that you can't explicitly re-run a search. Feels like if you hit the run button again it should do a complete new search.

### Error display

- [x] We currently just display a generic error message when the model is incorrect or the there is no key for the model. We either need a non-generic string or to display the exception message (probably preferable to start).

### Topic to discuss (not necessarily blocking)
- [ ] ~~**Progress UX for super-large trajectories.** Searched a 371M-token trajectory; progress was poor and final result appeared empty. May not be realistic to bite off in this PR.~~
- [x] Larger transcripts get chunked into segments, and the explanations just get [concatenated with "[Segment n]" headings](https://github.com/meridianlabs-ai/inspect_scout/blob/77c30dd752478477151263eba6ac9e3afec07d69/src/inspect_scout/_llm_scanner/_reducer.py#L30-L36). Only the `value` goes through the LLM reducer. Let's show `value` first, and then we'll have a second _Explanation_ heading before the `explanation`.

  Screenshot 2026-04-30 at 8 49 57 PM

packages/react/src/components/markdownRendering.ts
brandly: it's just down below in `markdownRenderers`!

https://github.com/meridianlabs-ai/ts-mono/pull/155/changes/BASE..8bfdd8fcdd8d3cdd7b91b6eaef6d542bcc3ec561#diff-a41e0962b02221c2ecbaee195f36be9eb23709d8a41dd0fd2bda590e6b0a6e6fR301-R305

dragonstyle: 💀

PR #163, opened 2026-04-30
Put test files alongside the code being tested

brandly:
I saw a comment about this for some new tests. We mostly do this across the repo, but the newly added `apps/inspect/` has a separate `tests/` directory.

This PR moves files out of `tests/` so that they site alongside the code being tested. The first commit is for `__tests__/markdownSecurity.test.ts`. It was the only file in a directory like that, so potentially it was intentional and I should move it back.


apps/inspect/src/tests/README.md
brandly, line 1: Here's Claude's justification for removing this file, which seems sound:

>   Already covered:
>   - "Test observable behavior, not implementation" → root AGENTS.md
>   - "Tests must be isolated/deterministic" → root AGENTS.md
>   - "Prefer real objects over mocks" → root AGENTS.md
>   - File extension hint (.test.ts vs .test.tsx) → ts-mono CLAUDE.md
>   - Per-package pnpm test and pnpm test:watch are obvious from package.json
>   - MSW / hook integration testing patterns → apps/scout/design/front-end-testing.md
> 
>   Actually worth salvaging from the deleted README:
>   - The mocking advice (vi.fn(), vi.mock()) is the most concrete bit — but it's also somewhat contradicted by front-end-testing.md which warns that vi.mock() couples
>   tests to implementation. So that's stale guidance, not advice we'd want to keep.
>   - test:coverage was listed but the script doesn't exist — that line was already wrong.

epatey (approved):

PR #171, opened 2026-05-01
viewer: show all scorer columns by default + tune samples-grid layout

### LINKED ISSUE #170: Feat: Show all scores by default; allow scrolling off screen

dragonstyle:
Fixes #170

## Summary

- Default the samples list to **showing every scorer column** instead of just the first. `useSelectedScores` now derives all available scorers when the user has no explicit selection; toggles in the column chooser still persist.
- **Tune the samples-grid sizing** so the new column count fits cleanly:
  - Scorer columns are sized to their header text (~6.2px/char + 24px), capped at 120px. Score values are tiny so the header is the binding constraint; long names like `exact_match_strict` no longer truncate, short ones stay compact.
  - Body columns get higher `minWidth` floors (input 240, target 120, answer 120) so they remain readable when many scorer columns share the row.
- **`SampleSummaryView` for many scores:** 1-3 scores still render inline alongside `Id` / `Input` / `Target` / `Answer`; 4+ scores move into a dedicated wrapping band underneath that uses `repeat(auto-fill, minmax(7em, 1fr))` and falls back to vertical scroll for the 30+ case.
- **Remove startup wobble** that was distracting on the same screens:
  - Bake column visibility into `initialState.columnVisibility.hiddenColIds` so ag-grid applies hides synchronously on first paint, instead of mounting all-visible and then redistributing flex columns from `onGridReady`.
  - Set `suppressColumnMoveAnimation: true` on both `SamplesGrid` and `LogListGrid` (Tasks / Folders / Samples-Panel) so programmatic visibility changes don't trigger ag-grid's 0.2s `left`/`width` cell transition.
  - Seed `epoch` visibility from `epochs > 1` in the Samples tab predicate so it no longer flashes visible → hidden once the seed lands in the store.

## Test plan

- [x] `pnpm check` (manypkg + lint + typecheck + format:check) green
- [x] `pnpm test` — 437 unit tests across inspect, scout, react packages
- [x] `pnpm e2e` — 109 Playwright tests across inspect + scout
- [x] Multi-scorer eval: confirm all scorer columns are visible on first load with no growing/sliding animation
- [x] 5+ scorer eval: scorer columns wrap appropriately; input column stays usable
- [x] Single-sample summary with 1-3 scores: inline layout (unchanged)
- [x] Single-sample summary with 4+ scores: wrapping band beneath the main grid; band scrolls when row count gets extreme
- [x] Tasks / Folders / Samples-Panel views: no column-resize animation on first paint


dragonstyle:
Also fixes #172

PR #176, opened 2026-05-04
Bidirectional column ↔ text filter sync for SampleList

dragonstyle:
## Summary

Reworks the SampleList filtering UX so the toolbar text filter and the
ag-grid column-header filters describe the same logical narrowing and
stay in sync as the user edits either surface.

- **All visible single-log columns are now expressible in the filter
  DSL.** Adds variables (`tokens`, `duration`, `retries`, `limit`,
  `has_limit`, `completed`, `input`, `target`, `error`, `answer`) and
  an `answer_contains()` function. `filterSamples` / `filterExpression`
  now take a `SamplesDescriptor` so `answer` can resolve through the
  selected scorer.
- **Bidirectional sync.** Column-header filter changes synthesize a
  filtrex expression and write it to the toolbar; toolbar edits parse
  via a small AST recognizer (recursive-descent over the existing
  CodeMirror tokenizer) and apply a `FilterModel` to the grid.
- **Round-trippable subset only.** The recognizer accepts a top-level
  conjunction of single-column predicates; anything outside that
  surface (OR, arithmetic, mixed types, regex metachars in unsupported
  positions) is treated as expression-only.
- **Expression-only mode.** When the typed text isn't round-trippable,
  the column-header filter buttons are hidden so the user can't
  accidentally overwrite their typed expression.
- **Score columns pick a filter type by `scoreType`** — numeric scores
  keep the number filter; categorical / pass-fail use the text filter.
  Boolean / object / list / "other" scores aren't sync'd at all
  (filtrex sees the underlying value, the column UI sees a stringified
  version, so they wouldn't match).
- **Sample ID isn't sync'd** for the same reason — `sample.id` may be
  numeric or string and filtrex's `==` is strict.
- **Round-trip stability fixes** along the way:
  - blank/notBlank uses filtrex's `None` constant (was `null`, which
    isn't a registered filtrex constant).
  - Empty model clears text vs. partial model leaves text alone (the
    early returns previously collapsed both into "leave text alone").
  - Partial synthesis is refused; the text→model effect preserves
    column entries the synthesizer would have skipped.
  - Same-column OR is dropped from synthesis (recognizer only handles
    AND; emitting OR would have round-tripped through the recognizer
    to `null` and cleared the grid).
  - Regex metachars use single-element character classes (`a[.]b`)
    since filtrex's lexer rejects bare `\.`.
  - Tokenizer's STRING regex now matches filtrex's actual escape
    rules (`\"` and `\\`); synthesizer uses minimal escapes so it
    doesn't emit JSON-style `\n` etc. that filtrex rejects.

Bonus UX: bumped score-column width budget so the active-filter icon
has room next to the sort icon under the Balham theme; added a 4 px
right gutter to all sample-grid header cells so funnel icons aren't
flush against the column separator.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm lint` clean
- [x] `pnpm test` — 230 passing (covering parser, recognizer,
      synthesizer, and end-to-end text→model→text round-trip stability,
      including filter values with quotes / backslashes / regex
      metachars)
- [x] `pnpm build` clean
- [x] Manual smoke (UI reviewer)
  - [x] Type `tokens > 50`; the Tokens column header funnel lights up.
  - [x] Add a Duration column filter via the header; text becomes
        `tokens > 50 and duration > 1`, both funnels stay active.
  - [x] Clear the last column filter; the text clears.
  - [x] Type a non-round-trippable expression
        (`score > 5 or score < 1`); column-header filter buttons hide
        and the typed text is preserved.
  - [x] Filter a Target column with a value containing a quote or
        regex metacharacter; round-trips cleanly through the toolbar.

### DIFF
--- a/apps/inspect/src/app/log-view/tabs/SamplesTab.tsx
+++ b/apps/inspect/src/app/log-view/tabs/SamplesTab.tsx
@@ -1,4 +1,4 @@
-import type { ColDef } from "ag-grid-community";
+import type { ColDef, GridApi } from "ag-grid-community";
 import { AgGridReact } from "ag-grid-react";
 import {
   FC,
@@ -32,6 +32,8 @@ import {
 import { useStore } from "../../../state/store.ts";
 import { ApplicationIcons } from "../../appearance/icons.ts";
 import { NavbarButton } from "../../navbar/NavbarButton.tsx";
+import { filterModelToText } from "../../samples/sample-tools/filterModelToText.ts";
+import { buildSampleFilterRegistry } from "../../samples/sample-tools/filterRegistry.ts";
 import { ColumnSelectorPopover } from "../../shared/ColumnSelectorPopover.tsx";
 import { getFieldKey } from "../../shared/gridUtils.ts";
 import {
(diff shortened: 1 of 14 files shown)


dragonstyle:
fixes #177

PR #185, opened 2026-05-05
Fetch pending sample data directly from S3 in viewer

rasmusfaber:
The viewer currently proxies every segment zip for live samples through the inspect view server, which bottlenecks on server CPU and network for samples with many segments. This PR teaches the client to call a new presigned-URL endpoint, fetch segment zips directly from S3 in parallel, and chunk requests via a server-driven `has_more` cursor so the renderer paints between chunks instead of starving on a single long sweep. Any failure (presign unavailable, old server, network error, non-S3 buffer) falls back transparently to the proxy endpoint.

Server side lives in the parent inspect_ai repo (UKGovernmentBEIS/inspect_ai#3835).

### DIFF
--- a/apps/inspect/src/client/api/client-api.test.ts
+++ b/apps/inspect/src/client/api/client-api.test.ts
@@ -1,0 +1,128 @@
+import { describe, expect, test, vi } from "vitest";
+
+import { clientApi } from "./client-api";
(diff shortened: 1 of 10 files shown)


rasmusfaber:
(The linting and typechecking also fails on main)

dragonstyle (approved):

PR #187, opened 2026-05-05
fix: cmd+click/middle-click opens log in background tab

revmischa:
## Summary

- Render the task column in the log list grid as a real `<a>` tag so cmd+click uses native browser background-tab behavior instead of `window.open()` (which always steals focus)
- Skip modifier-key clicks in `onRowClicked` handler to avoid double navigation
- Fix `window.open()` URLs to include `window.location.pathname` for correct hash routing when the viewer is embedded at a subpath (e.g. `/eval-set/xxx#/logs/`)
- Add `openInNewTab` helper for keyboard shortcuts and other non-mouse contexts

https://github.com/user-attachments/assets/117550f2-20b0-408b-a1eb-e196c3eb178d

### DIFF
--- a/apps/inspect/src/app/log-list/grid/LogListGrid.tsx
+++ b/apps/inspect/src/app/log-list/grid/LogListGrid.tsx
@@ -31,6 +31,7 @@ import { useStore } from "../../../state/store";
 
 import "../../shared/agGrid";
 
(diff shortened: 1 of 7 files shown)


dragonstyle:
Thanks opened a pr with formatting and type fixes

PR #191, opened 2026-05-06
Fix runtime errors when loading viewer libs in browsers

rasmusfaber:
## Summary

Consumers of `@meridianlabs/log-viewer` and `@meridianlabs/inspect-scout-viewer` hit two runtime errors in their browsers: `__require is not defined` whenever either viewer loads under Vite's dev server (so `pnpm dev` doesn't work — consumers fall back to `pnpm build && pnpm preview`), and `TypeError: Cannot set property Package` when an app loads both viewers on the same page. Both stem from the libs shipping things they shouldn't — bundled CJS shims for `react/jsx-runtime`, plus a private 6 MB MathJax copy embedded in each lib.

The fix is to externalize `react` / `react-dom` subpaths (was only externalizing the bare names, so Rolldown bundled the subpaths as CJS) and to externalize `mathjax-full` + `markdown-it-mathjax3` as peerDependencies. Each lib drops ~6 MB.

## Notes for reviewers

- **Breaking for consumers** of the published packages: the MathJax peerDeps require any consuming app to add `mathjax-full` and `markdown-it-mathjax3` to its own dependencies.

### DIFF
--- a/apps/inspect/package.json
+++ b/apps/inspect/package.json
@@ -72,6 +72,8 @@
     "eslint-plugin-react-hooks": "^7.1.1",
     "fake-indexeddb": "^6.2.2",
     "globals": "^17.5.0",
+    "markdown-it-mathjax3": "^5.2.0",
+    "mathjax-full": "^3.2.2",
     "msw": "^2.13.6",
     "picocolors": "^1.1.1",
     "postcss": "^8.5.12",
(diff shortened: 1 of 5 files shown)


dragonstyle:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #203, opened 2026-05-06
Make find work everywhere in the sample viewer

rasmusfaber:
## Summary

The sample viewer's find experience was broken in several related ways across all four sample tabs. 

I have tried a balance that avoids a full rewrite, but also doesn't just add duct-tape everywhere.

So we rebuild it around a single sample-wide search source for the transcript, fix the chat-tab highlight loss at its DOM-mutation root, and remove the early-bail that left Scoring/Metadata/JSON entirely unsearchable.

The bugs that drove this work:
- **Cross-row navigation didn't work** in the transcript — matches in collapsed sub-agent rows were invisible to the counter and unreachable via Next/Prev.
- **Counter inflated by unrendered events** (state/store etc.) so users saw "1 of 20" but could only navigate to 10 of them.
- **Highlights vanished after ~1 second** on the Messages tab — caused by Virtuoso re-mounting every row when `components.Item` was a fresh function reference each render.
- **Search did nothing on Scoring/Metadata/JSON** because FindBand bailed out when `countAllMatches` returned 0; those tabs are static markup with no registered counter.
- **Reasoning matches were missed** when the model redacted the raw `reasoning` field — the search indexed the encrypted blob, not the visible summary.

## What changed

- `useTranscriptSearchSource` + `sampleSearch`: walks all events in the sample, knows which swimlane row each match lives in, switches rows + scrolls + auto-expands when navigating across rows, and skips events whose row isn't surfaced in the rendered tree (via the same `hiddenEventTypes` the renderer uses) so the counter only includes reachable matches.
- `FindTargetContext`: replaces imperative DOM-mutation with declarative auto-expand. `ExpandablePanel` self-detects when its subtree contains the active term and expands; mounts optimistically so Virtuoso re-renders during search don't flash collapsed→expanded.
- Hoisted `ChatVirtuosoItem` to module scope and memoized `LiveVirtualList`'s `Footer`/components — fixes the "highlight disappears after 1 second" symptom by stopping Virtuoso's mass-row remount.
- `messageSearchText` and `eventText` reasoning case: mirror the renderer's redacted→summary fallback so search hits what the user actually sees.
- FindBand no longer bails on `total === 0`; runs the find anyway and tracks "no results" as separate state. Static-content tabs (Scoring, Metadata, JSON) now search via `window.find` with no counter shown.

## Test plan

- [x] Sample with sub-agent rows: search a term that appears across multiple rows, verify counter, Next/Prev, and that target panels auto-expand
- [x] Search inside a clipped `ExpandablePanel` — confirm it expands without flashing collapsed→expanded
- [x] Messages tab: redacted-reasoning summary search (e.g. `hiccup` in the test eval) finds the term and the highlight stays past the ~1s mark
- [x] Scoring / Metadata / JSON tab: typing a visible term highlights it (no "X of Y" counter, that's expected); typing a non-existent term shows "No results"
- [x] Transcript Next/Prev: swimlane bar collapses on Next and expands on Prev (matches manual scroll-down/up)
- [x] Match counter accuracy on transcript: counts only reachable matches (not state/store events)

### DIFF
--- a/apps/inspect/src/app/log-view/LogViewLayout.tsx
+++ b/apps/inspect/src/app/log-view/LogViewLayout.tsx
@@ -1,7 +1,11 @@
 import clsx from "clsx";
 import { FC, useEffect, useRef } from "react";
 
-import { ErrorPanel, ExtendedFindProvider } from "@tsmono/react/components";
+import {
+  ErrorPanel,
+  ExtendedFindProvider,
+  FindTargetProvider,
+} from "@tsmono/react/components";
 
 import { ActivityBar } from "../../components/ActivityBar";
 import { FindBand } from "../../components/FindBand";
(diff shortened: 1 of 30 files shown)


dragonstyle:
Thank you! Closed in favor of rebased #211

PR #231, opened 2026-05-18
Smaller Scout bundles

brandly:
I saw that these bundles were really big, and I told Claude to investigate.

Initial-load bundle (CSS + main JS — what the user pays for on first paint)

```
┌────────────┬───────────┬──────────┬──────────────────┐
│            │  Before   │  After   │        Δ         │
├────────────┼───────────┼──────────┼──────────────────┤
│ main JS    │ 3,191 KB  │ 2,669 KB │ −522 KB (−16%)   │
├────────────┼───────────┼──────────┼──────────────────┤
│ CSS        │ 1,108 KB  │ 510 KB   │ −598 KB (−54%)   │
├────────────┼───────────┼──────────┼──────────────────┤
│ Total      │ 4,299 KB  │ 3,179 KB │ −1,120 KB (−26%) │
├────────────┼───────────┼──────────┼──────────────────┤
│ gzip total │ ~1,428 KB │ ~879 KB  │ −549 KB (−38%)   │
└────────────┴───────────┴──────────┴──────────────────┘
```

The big drop in CSS size is that with `maxSize: Infinity` in the `postcss-url` means any `url()` reference would get base64-embedded in the CSS file including the bootstrap-icons .woff2 (~130 KB) and codicon .ttf (~120 KB).


brandly:
> dropping apache-arrow removed the transitive @types/node@24.12.4 entry from the lockfile. That older version was visible to inspect-components through the util → apache-arrow → @types/node chain. With apache-arrow gone, only @types/node@25.7.0 remains (in apps/scout/inspect), and it's not reachable from inspect-components.

If I manually add that package and try to run a `pnpm install`, I get

```
 ERR_PNPM_NO_MATURE_MATCHING_VERSION  Version 25.7.0 (released 6 days ago) of @types/node does not meet the minimumReleaseAge constraint

This error happened while installing a direct dependency of /Users/brandly/code/inspect_scout/src/inspect_scout/_view/ts-mono/apps/scout
```

So essentially, we already had newer-than-a-week-old dependencies before setting the `minimumReleaseAge`. Tomorrow I can run an install and get this passing CI.

packages/util/src/arrow.ts
```diff
@@ -22,20 +17,13 @@ export const decodeArrowBase64 = (base64: string) => {
...
-  // the wire
-  const arrowTable = tableFromIPC(bytes);
-
-  // Convert to uncompressed Arrow IPC format
-  const uncompressedBytes = tableToIPC(arrowTable);
-
-  // Now Arquero can read the uncompressed Arrow data
-  let table = fromArrow(uncompressedBytes);
+  const input = bytes instanceof Uint8Array ? bytes : new Uint8Array(bytes);
+  let table = fromArrow(input);
```
brandly, line 26: this is the piece to flag and verify further!

brandly: @dragonstyle added a test here and manually verified that I'm able to view a scan. seems like we're fine without `apache-arrow`!

Screenshot 2026-05-20 at 1 15 50 PM

dragonstyle:
Will you give this a try in vscode? I'm suspicious that CSP might cause issues there...

brandly:
@dragonstyle after getting sidetracked by https://github.com/meridianlabs-ai/inspect_vscode/pull/87, I'm now able to confirm that the `asciinema-player` lazy loading doesn't cause any problems in vscode. I think it's all working well but let me know if there's anything else I should confirm!

PR #235, opened 2026-05-19
Add provisional checkpointing UX to the viewer.

epatey:
> [!CAUTION]
> Do not merge until https://github.com/UKGovernmentBEIS/inspect_ai/pull/3940 is ready to be merged immediately after.

This PR has no impact in the abscense of an eval with checkpoints.

### DIFF
--- a/packages/inspect-components/src/transcript/CheckpointEventView.tsx
+++ b/packages/inspect-components/src/transcript/CheckpointEventView.tsx
@@ -1,0 +1,105 @@
+import { FC, ReactNode } from "react";
+
+import type { CheckpointEvent } from "@tsmono/inspect-common/types";
+import { LabeledValue } from "@tsmono/react/components";
+import { formatBytes, formatMs } from "@tsmono/util";
+
+import styles from "./CheckpointEventView.module.css";
+import { EventRow } from "./event/EventRow";
+import { formatTitle } from "./event/utils";
+import { TranscriptIcons } from "./icons";
+import { EventNode } from "./types";
+
+interface CheckpointEventViewProps {
+  eventNode: EventNode<CheckpointEvent>;
+  className?: string;
+}
+
+const SizeAndDuration: FC<{ size_bytes: number; duration_ms: number }> = ({
+  size_bytes,
+  duration_ms,
+}) => (
+  <>
+    <LabeledValue label="Size" layout="row">
+      {formatBytes(size_bytes)}
+    </LabeledValue>
+    <LabeledValue label="Duration" layout="row">
+      {formatMs(duration_ms)}
+    </LabeledValue>
+  </>
+);
+
+export const CheckpointEventView: FC<CheckpointEventViewProps> = ({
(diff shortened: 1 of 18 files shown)


packages/inspect-components/src/transcript/CheckpointEventView.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+// TODO: This could use formatTime, but that loses sub-second precision. Maybe that's fine
+const formatSeconds = (ms: number): string =>
+  `${(ms / 1000).toLocaleString(navigator.language, {
+    minimumFractionDigits: 0,
+    maximumFractionDigits: 3,
+  })} s`;
+
+// TODO: Is there really no helper that renders byte counts? e.g. 1.23kb, mb, gb, etc
+const formatBytes = (bytes: number): string => {
```
dragonstyle: This seems like a great utility. I, too, was surprised we didn't have one!

packages/inspect-components/src/transcript/CheckpointEventView.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { TranscriptIcons } from "./icons";
+import { EventNode } from "./types";
+
+interface CheckpointEventViewProps {
+  eventNode: EventNode<CheckpointEvent>;
+  className?: string;
+}
+
+// TODO: This could use formatTime, but that loses sub-second precision. Maybe that's fine
+const formatSeconds = (ms: number): string =>
```
dragonstyle: formatTime is used for things where sub second prevision likely doesn't matter (e.g. formatting eval duration). For this use, this seems entirely appropriate and perhaps could be made a utility for formatting ms output.

packages/inspect-components/src/transcript/CheckpointEventView.tsx
```diff
@@ -0,0 +1,132 @@
...
+    unitIdx++;
+  }
+  const formatted = value.toLocaleString(navigator.language, {
+    minimumFractionDigits: 0,
+    maximumFractionDigits: unitIdx === 0 ? 0 : 2,
+  });
+  return `${formatted} ${units[unitIdx]}`;
+};
+
+const Metric: FC<{ label: string; value: string }> = ({ label, value }) => (
```
dragonstyle: There is `LabeledValue` which I think would be a preferred approach (basically the same thing as `Metric` but with more opinionated formatting).

packages/inspect-components/src/transcript/CheckpointEventView.tsx
```diff
@@ -0,0 +1,132 @@
...
+              </div>
+            </div>
+          ))}
+        </div>
+      </div>
+    );
+  }
+
+  return (
+    <EventPanel
```
dragonstyle: `EventRow` is the alternative if a simple single line rendering is preferred.

dragonstyle (commented):
This looks good - a few suggestions.

PR #237, opened 2026-05-20
Synchronously check for single-file mode

### LINKED ISSUE #77: Inspect view: opening a single log should not fetch data for every other log in the directory
## Summary

We embed inspect view in our application as an iframe and only ever link to individual logs via the `?log_file=...` URL parameter . Our users don't see the full log-listing view. We run one Docker container with inspect view pointing at a log directory on disk, and the iframe is served from that container.

While using it this way, we noticed that loading a single log against a directory with ~50 .eval files produces ~800 of HTTP requests before the log renders. The request count scales linearly with directory size regardless of whether the user asked for one log or the full listing.

## Environment

- Inspect version: 0.3.200
- Python: 3.12

## Reproduction

1. Run inspect view --log-dir <dir> against a directory with several .eval log files.
2. Open the browser devtools Network tab.
3. Navigate directly to a single log via ?log_file=<path> (deep link, not the listing).
4. Observe request count climb fast, asking for every log, not only the one linked

## Desired behavior

When the view is opened via a deep link to a specific log, loading that log should not require fetching data for every other log in the directory. The cost of opening one log should be roughly constant, independent of directory size.

brandly:
Fixes #77

I had trouble truly reproducing this. When the `log_file` query param is present, we'll already go into single-file mode

https://github.com/meridianlabs-ai/ts-mono/blob/8025d3440770658c046a1fcb109fc28af47ca6a1/apps/inspect/src/app/App.tsx#L216-L219

But this is setting state within a `useEffect` call, so it seems possible for there to be a race condition that's resulting in this not always being the case. This value is not something that changes throughout the lifetime of the application, so we'll detect it synchronously on startup and prevent any races.

With a directory of 32 `.eval` files, we make just 8 fetch calls:
Screenshot 2026-05-20 at 12 36 24 PM

### DIFF
--- a/apps/inspect/src/app/App.tsx
+++ b/apps/inspect/src/app/App.tsx
@@ -80,10 +80,6 @@ export const App: FC<AppProps> = ({ api }) => {
   const loadLog = useStore((state) => state.logActions.syncLog);
   const pollLog = useStore((state) => state.logActions.pollLog);
 
-  const setSingleFileMode = useStore(
-    (state) => state.appActions.setSingleFileMode
-  );
-
   // Load a specific log
   useEffect(() => {
     const loadSpecificLog = async () => {
@@ -194,7 +190,6 @@ export const App: FC<AppProps> = ({ api }) => {
       if (embeddedState) {
         const state = JSON5.parse(embeddedState.textContent || "");
         onMessage({ data: state });
-        setSingleFileMode(true);
       } else {
         // For non-route URL params support (legacy)
         const urlParams = new URLSearchParams(window.location.search);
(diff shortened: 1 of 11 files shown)


apps/inspect/src/app/App.tsx
```diff
@@ -210,13 +206,12 @@ export const App: FC<AppProps> = ({ api }) => {
           setLogDir(undefined);
           // Load just the passed file
           setLogFiles([{ name: resolvedLogPath }]);
-          setSingleFileMode(true);
         } else {
           // If a log file was passed, select it
           const log_file = urlParams.get("log_file");
           if (log_file) {
+            await initLogDir();
```
brandly, line 213: This fixes an issue where the request to `/api/log-info` for the given eval would fail with a 403 due to passing the filename instead of the fully resolved file path to the endpoint. The 403 would flash on screen very briefly before the page loaded.

dragonstyle (approved):

PR #241, opened 2026-05-21
Theming

Aprillion:
Shared theme selector with Event Colors for inspect/scout/hawk viewers, discussion and screenshots in https://inspectcommunity.slack.com/archives/C0B265TQQAF/p1779346946869979?thread_ts=1779291310.781859&cid=C0B265TQQAF


tooling/vite-plugins/index.js
```diff
@@ -0,0 +1,43 @@
...
+export function inlineThemeBootstrap(
+  entry,
+  placeholder = "<!-- THEME_BOOTSTRAP -->"
+) {
+  return {
+    name: "inline-theme-bootstrap",
+    transformIndexHtml: {
+      order: "pre",
+      async handler(html) {
+        const result = await esbuild({
```
dragonstyle: Dev server runs a full esbuild({bundle: true, ...}) per HTML reload. Each dev nav/HMR HTML transform spends ~50-200ms re-bundling the same entry. Cache the output keyed on entry mtime, or build once at server-start?

packages/inspect-common/src/theme/bootstrap.ts
```diff
@@ -0,0 +1,378 @@
...
+    : name;
+  return { isDark, variant, base };
+};
+
+/**
+ * Read a persisted `themePreference` out of a zustand-persisted settings
+ * blob. Apps that have no preference store omit `storageKey` and always
+ * resolve to `system`.
+ */
+export const readThemePreference = (
```
dragonstyle: I think the storage and parsing of the result is more an application level concern. Perhaps rather than having this, the createApplyTheme options should include something like:

```typescript
{
  ...
  readPreference: () => readInspectThemePreference(),
  ...
}
```

and each app can implement this as they see fit.

packages/theme/src/bootstrap.ts
```diff
@@ -0,0 +1,378 @@
...
+ *    base.css (`body[class^="vscode-"]`). We deliberately set only
+ *    `vscode-dark` (never `vscode-light`) so light standalone stays on
+ *    pure Bootstrap — see the comment at the toggle call below.
+ *
+ * Nothing here self-executes; callers must invoke the returned function.
+ * That keeps the module import-safe for tests/SSR and lets the caller
+ * control timing (the inline <script> calls it before first paint).
+ */
+
+export type ThemePreference =
```
dragonstyle, line 18: I sort of think that ideally this wouldn't be in inspect-common, but would be in theme still. Perhaps with changes below it will not need to be here any longer?

apps/inspect/src/theme/themePreference.ts
```diff
@@ -0,0 +1,70 @@
...
+
+import { SETTINGS_STORAGE_KEY } from "./constants";
+
+// Inspect has no general settings store, so theme preference gets its own
+// tiny localStorage record. The shape (`{ state: { themePreference }, ... }`)
+// matches what @tsmono/inspect-common/theme/bootstrap's readThemePreference parses, so the
+// inline bootstrap and this hook share one source of truth.
+export { SETTINGS_STORAGE_KEY };
+
+const read = (): ThemePreference =>
```
dragonstyle: I think it would be nice to use a general userSettings approach like Scout does. I think we'll have more local storage based settings so it would nice to take a uniform approach and have that settings store here ready for new settings.

tooling/vite-plugins/index.js
```diff
@@ -0,0 +1,43 @@
...
+          entryPoints: [entry],
+          bundle: true,
+          format: "iife",
+          target: "es2020",
+          minify: true,
+          write: false,
+        });
+        const code = result.outputFiles[0].text.trim();
+        return html.replace(placeholder, `<script>${code}</script>`);
```
dragonstyle, line 60: I think this should throw if the placeholder is missing.

dragonstyle (changes requested):
Thanks for this and sorry for the slow review!

tooling/vite-plugins/index.js
Aprillion: error added

tooling/vite-plugins/index.js
Aprillion: now cached (I hope ... opus 4.8 needed little nudge to not over-cache rejected promise, though not sure if that can happen and/or how to test that branch ..I'm not completely up-to-date on esbuild lifecycle 🙈)

packages/inspect-common/src/theme/bootstrap.ts
Aprillion: ~moved to app layer, not sure if it's fine to duplicate the code for now (to make it easy to change in 1 app) or to share it but at different layer of abstraction?~

nevermind, moving as a util to restored `theme` package without duplication

apps/inspect/src/theme/themePreference.ts
Aprillion: ~hm, now the approach is so uniform that `readPreference.ts` are duplicated in the 2 apps 🤔 ..it's optimized for independent change and agentic reviewers were happy to keep it even if I am not quite sure myself ... but if we decide to go the `@tsmono/theme` way to solve the architecture dilemma from previous comment, we could share the code there~

moved to `@tsmono/theme`

apps/inspect/package.json
```diff
@@ -93,7 +93,6 @@
     "@tsmono/inspect-components": "workspace:*",
     "@tsmono/react": "workspace:*",
     "@tsmono/scout-components": "workspace:*",
-    "@tsmono/theme": "workspace:*",
```
Aprillion, line 96: oh wait, theme package existed on main already? oops, let me restore it 🙈

packages/theme/src/bootstrap.ts
Aprillion: oh, `@tsmono/theme` package already existed on main => restoring it and moving code there where it belongs 👍

apps/inspect/package.json
Aprillion: ok, shared code now lives in the correct package🤞

dragonstyle:
2 notes from Codex:

Low: base.css (line 257) appears to have dropped the prior VS Code-only --bs-border-radius: 0 and --bs-border-radius-lg: 0 overrides. If the rounded controls in VS Code are intentional, fine; otherwise this is a small visual regression from current main.

Low/design: bootstrap.ts (line 221) still uses one module-level currentApplyTheme and one window.__APPLY_BROWSER_THEME__. That is acceptable for one viewer per page, but if Hawk or another host ever embeds multiple viewers in the same window, they will overwrite each other’s global apply function. I would either document that assumption or namespace the global per app.

Aprillion:
> Low: base.css (line 257) appears to have dropped the prior VS Code-only --bs-border-radius: 0 and --bs-border-radius-lg: 0 overrides. If the rounded controls in VS Code are intentional, fine; otherwise this is a small visual regression from current main.

intentional because I think it looks better, but let me know if you want to go back to the sharp edges (not committed yet):
|currently rounded|if reverted to angular|
|---|---|
|image|image|

> Low/design: bootstrap.ts (line 221) still uses one module-level currentApplyTheme and one window.**APPLY_BROWSER_THEME**. That is acceptable for one viewer per page, but if Hawk or another host ever embeds multiple viewers in the same window, they will overwrite each other’s global apply function. I would either document that assumption or namespace the global per app.

obviously the theme is intentionally shared in hawk,, but good idea to add a comment (PR updated) 😅

dragonstyle:
TY!

That `border-radius` change will change the radii across the whole app (other cards, etc). We set that to 0 to make it look more vscode native (since it uses no border radius) - I think we should leave that as is for vscode and then I'll merge this!

Aprillion:
> TY!
> 
> That `border-radius` change will change the radii across the whole app (other cards, etc). We set that to 0 to make it look more vscode native (since it uses no border radius) - I think we should leave that as is for vscode and then I'll merge this!

eeh, but native vscode uses 4px border radius for most buttons and stuff these days, not 0 - https://github.com/microsoft/vscode/pull/280457

dragonstyle (approved):

PR #244, opened 2026-05-22
Log editing UI: tag + metadata editor dialogs and supporting components

ransomr:
## Summary

Implements the viewer-side UI for the log editing feature (tags + metadata) that lands in the inspect_ai repo. Adds the edit dialogs, the chip strip / overflow logic, the change summary, the ETag plumbing for concurrent-modification protection, and the VS Code bridge methods.

## Highlights

- **Tag editor**: `TagsField`, `TagStrip`, `TagChip`, `EditTagsDialog`. Chips collapse to two rows + `…` overflow pill on narrow rows; clicking any chip or pill opens the dialog. Pills use `text-size-smaller` and a flat outline-only look that matches the header.
- **Metadata editor**: `EditMetadataDialog` with a single CSS grid (subgrid rows) so key / value / delete columns align across rows. Borderless flat value editor; `+ N` change summary in the footer.
- **Provenance fields**: shared `ProvenanceFields` component used by both dialogs.
- **Error formatting**: `formatEditError` + `unwrapFastapiDetail` so 400 / 409 / 412 responses render the server's actionable text (not the raw `{"detail": …}` envelope), and fetch network errors render as `Connection lost — view server unreachable`.
- **Concurrent-modification protection**: a per-log ETag cache in `clientApi`, seeded from `LogDetails.etag` (lifted off the `get_log_info` head_object response) and refreshed from each successful edit, so the first AND subsequent saves carry `If-Match`.
- **VS Code bridge**: `edit_log` / `get_user_info` JSON-RPC methods with `MethodNotFound` fallback and FastAPI-style status-code → `ApiError` mapping.
- **Theme**: VS Code overrides for `.btn-secondary`, `.modal` borders, `.form-control` background / border / placeholder, and a new `--inspect-hint-foreground` so the italic "optional" hint is readable in dark themes.

## Test plan

- [x] `pnpm --filter inspect typecheck`
- [x] `pnpm --filter inspect test --run` (28 files / 389 tests)
- [ ] Verify chip collapse / overflow visually at narrow widths
- [ ] Verify edit dialogs render correctly in VS Code dark theme

This PR pairs with the corresponding inspect-ai PR which lands the server-side endpoints and `inspect_ai.log._edit` API.


apps/inspect/src/app/log-view/title-view/AutogrowText.tsx
```diff
@@ -0,0 +1,75 @@
...
+  // (e.g. the EditMetadataDialog has a scroll wrapper around the
+  // whole metadata table).
+  maxHeightPx?: number;
+}
+
+// A single-line-by-default textarea that grows with its content. By
+// default there's no upper bound on height — the surrounding container
+// handles scrolling — so a long JSON value renders at full height
+// rather than embedding its own inner scrollbar.
+export const AutogrowText: FC<AutogrowTextProps> = ({
```
dragonstyle, line 22: This seems like a useful general component - there is a potentially similar thing used in the scout search panel (capped to a certain number of lines as user types). @brandly will know which component (if any) that is.

apps/inspect/src/app/log-view/title-view/EditMetadataDialog.tsx
```diff
@@ -0,0 +1,517 @@
...
+  const [newKey, setNewKey] = useState("");
+  const [newType, setNewType] = useState<NewType>("string");
+  const [author, setAuthor] = useState("");
+  const [reason, setReason] = useState("");
+  const [submitting, setSubmitting] = useState(false);
+  const [error, setError] = useState<string | undefined>();
+  // Captures the most recently added key so a layout effect can scroll
+  // the new row into view and focus its value textarea once it's in
+  // the DOM. Reset to null after handling.
+  const [pendingFocusKey, setPendingFocusKey] = useState<string | null>(null);
```
dragonstyle, line 171: If a user is in Vscode, opens a log file, opens the metadata dialog and then in VSCode switches to another editor tab and back, the whole app is re-mounted (we unmount the browser when our tab is backgrounded to reduce memory usage). This will cause anything that is `useState` to lose state. If you use `useStore` (adding needed props and setters), we persist the store and reload it automatically when the tab is restored.

I think this could be ok if the idea is that the dialog won't stay showing when backgrounding, but if the dialog stays open when the user resumes, I think we should preserve their values.

apps/inspect/src/app/log-view/title-view/EditTagsDialog.tsx
```diff
@@ -0,0 +1,262 @@
...
+  onSaved,
+}) => {
+  const api = useStore((state) => state.api);
+
+  const [tags, setTags] = useState<string[]>(currentTags);
+  const [pending, setPending] = useState("");
+  const [author, setAuthor] = useState("");
+  const [reason, setReason] = useState("");
+  const [submitting, setSubmitting] = useState(false);
+  const [error, setError] = useState<string | undefined>();
```
dragonstyle, line 39: Same comment as Metadata dialog.

apps/inspect/src/app/log-view/title-view/EditMetadataDialog.tsx
ransomr: Currently the dialog does not stay open on navigation. I'd like to leave as is for now - we can add later if needed.

apps/inspect/src/app/log-view/title-view/AutogrowText.tsx
ransomr: The Scout analog is `apps/scout/src/app/utils/autosizeTextarea.ts` (a ~30-line helper, not a component), wired up at `apps/scout/src/app/transcript/SearchPanel.tsx:340-349`:

```ts
// utils/autosizeTextarea.ts
export type AutosizeTextareaConfig = { minRows: number; maxRows: number };
export function autosizeTextarea(el: HTMLElement, config: AutosizeTextareaConfig): void {
  // computes line-height/padding/border, resets height to "auto",
  // reads scrollHeight, clamps between minRows*lh and maxRows*lh,
  // sets el.style.height and toggles overflow-y: auto at the cap.
}
```

```ts
// SearchPanel.tsx — call site
const SEARCH_TEXTAREA_AUTOSIZE: AutosizeTextareaConfig = { minRows: 2, maxRows: 10 };
const searchTextareaRef = useRef<HTMLElement | null>(null);
useLayoutEffect(() => {
  const inner = getInnerTextarea(searchTextareaRef.current);
  if (inner !== null) autosizeTextarea(inner, SEARCH_TEXTAREA_AUTOSIZE);
}, [query]);
```

Same job as `AutogrowText` (reset `height` → read `scrollHeight` → clamp → toggle inner scroll once capped) but different in three ways:

| | Scout `autosizeTextarea` | `AutogrowText` (this PR) |
|---|---|---|
| Shape | Function on an existing `HTMLElement` | React component that renders the `<textarea>` |
| Resize trigger | Caller's `useLayoutEffect` on value changes | Internal `ResizeObserver` |
| Cap unit | `minRows` / `maxRows` (computes line-height) | `maxHeightPx` (raw px), no min |
| Reset before measure | `el.style.height = "auto"` | `el.style.height = "0px"` |

Scout's helper is function-shaped because its textarea lives inside a shadow DOM web component (`el?.shadowRoot?.querySelector("textarea")`), so a self-rendering React component can't reach it.

### Suggested follow-up

Promote a single `autosizeTextarea(element, config)` helper into a shared package (`packages/inspect-components` or `packages/util`):

1. Reconcile config shape: pick one of `{minRows, maxRows}` vs `{minHeightPx?, maxHeightPx?}`, or accept both and normalize internally. Rows are friendlier (line-height stays correct under font-size changes), so probably canonicalize on rows and have a `pxFromRows` helper if a px API is still wanted.
2. Reconcile the height reset: `"0px"` vs `"auto"`. Both work; the `"0px"` variant was added in this PR because `scrollHeight` clamps to the previously-applied height in some Safari versions when the element re-mounts behind a hidden Modal. Worth a small test before standardizing.
3. Refactor `AutogrowText` to a thin React wrapper that:
   - renders `<textarea ref={ref} rows={minRows} ... />`,
   - runs the shared helper inside its own `ResizeObserver` (preserves the "auto" behavior in EditMetadataDialog where the value mounts inside a hidden modal),
   - exposes `minRows`/`maxRows` props.
4. Switch `SearchPanel` over to call the shared helper directly (the shadow-DOM consumer still can't take a self-rendering component).
5. Drop `apps/scout/src/app/utils/autosizeTextarea.ts` once the import sites are migrated.

Risk: SearchPanel currently measures via `getComputedStyle` on the inner shadow-DOM textarea — keep that path intact since shadow-rooted styles read fine but the *containing* element doesn't have the textarea's font/line-height. The shared helper takes the element to measure as its first argument, so this just works.

ransomr: See investigation above - this refactor makes sense to me, but I'd like to do as a separate PR.

dragonstyle (approved):

PR #246, opened 2026-05-22
Inspect - Move `api` to context

brandly:
This closely mimics how Scout handles things. The `<App>` takes an `api` prop, so that anyone can inject in their own `ClientAPI` implementation. We put the provided `api` in a context, and it can be accessed with `useApi()`.

One nice thing is that this reduces a bunch of uncertainty and `api?` checks. `useApi` will always return a `ClientAPI`.

`LogSlice` has methods that would grab the `api` from the store and then use various methods on it. Ultimately it would be good to disentangle this and separate network activity from the local store, but for now, we'll pass our `api` into `createLogSlice`. We follow this pattern in a few places for the store.

### DIFF
--- a/apps/inspect/src/components/DownloadButton.tsx
+++ b/apps/inspect/src/components/DownloadButton.tsx
@@ -1,6 +1,6 @@
 import { FC } from "react";
 
-import api from "../client/api/index";
+import { useApi } from "../state/store";
 
 import "./DownloadButton.css";
 
@@ -15,6 +15,7 @@ export const DownloadButton: FC<DownloadButtonProps> = ({
   fileName,
   fileContents,
 }) => {
+  const api = useApi();
   return (
     <button
       className={"btn btn-outline-primary download-button"}
--- a/apps/inspect/src/state/samplePollingInstance.ts
+++ b/apps/inspect/src/state/samplePollingInstance.ts
@@ -1,7 +1,19 @@
+import { ClientAPI } from "../client/api/types";
+
 import { createSamplePolling } from "./samplePolling";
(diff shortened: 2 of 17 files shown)


apps/inspect/src/components/DownloadButton.tsx
```diff
@@ -1,6 +1,6 @@
 import { FC } from "react";
 
-import api from "../client/api/index";
+import { useApi } from "../state/store";
```
brandly, line 3: it looks like in this case, we weren't respecting the api from the store, but now we'll read from the context

apps/inspect/src/state/samplePollingInstance.ts
```diff
@@ -1,7 +1,19 @@
...
+let injectedApi: ClientAPI | null = null;
+
+/**
+ * Records the api that the polling singleton should use. Called from
+ * initializeStore so the singleton stays consistent with the api the
+ * consumer passed into the App tree.
+ */
+export function setSamplePollingApi(api: ClientAPI) {
+  injectedApi = api;
+}
```
brandly, line 16: this file is the weirdest change and worth reviewers' attention

dragonstyle: This actually seems ok to me if we view it as 'initSamplePolling` - not totally unexpected that if we want a singleton to handle polling we might need this type of thing. 

> Aside: I wonder sample polling is trying to do a lot that react query would help with (repolling on an interval etc.) and if the longer term fix involves porting to react query.

dragonstyle (approved):
I think this looks ok...

PR #248, opened 2026-05-23
Consolidate duplicate code

brandly:
Both Scout and Inspect's viewers had identical `AppErrorBoundary` components and `useBreadcrumbTruncation` hooks. There was also a lot of overlap between their icons.

Surprisingly a couple icons were defined differently for each app, and some icons that were only defined for a specific app have been left up in app code.


dragonstyle (approved):
Nice!

PR #255, opened 2026-05-27
Grouped column headers for dict-returning metrics

kaifronsdal:
## Summary

Adds grouped column headers to the score views when an `EvalMetric` carries a `group` field (set by inspect_ai for dict-returning metrics like `frequency()` — see companion PR).

- **`ScoreGrid`** (headline table): optional second `<thead>` row with colspan group cells when ≥2 consecutive metrics share a non-null `group`.
- **`ScoreAgGrid`** (Scoring Detail modal): restructured from one sparse union grid into one AG Grid per `scoreGroup` (scorers sharing a `(group, name)` metric signature). Each grid gets its own `ColGroupDef` nesting and sizes to its own column count — fixes the pre-existing sparse-matrix layout.
- `groupScorers`/`metricsKey` now key on `(group, name)` so scorers with different metrics but coincident leaf labels don't merge.
- Index-based row fields, `suppressFieldDotNotation`, container padding, auto-height min unset, last-column header right-alignment fix, scoped `thead` reset.
- Regenerated `inspect-common` types for `EvalMetric.group`, `EvalScorer.value_schema`, `CategoricalSchema`.

Backward-compatible: logs without `group` render exactly as before.

## Companion

Depends on inspect_ai categorical-scoring PR (`feature/categorical-scores`) which populates `EvalMetric.group`. This PR is harmless without it (group is `undefined` → flat rendering).

## Test plan

- [x] `pnpm --filter inspect typecheck` / `lint` / `format`
- [x] `pnpm --filter inspect test` (350 pass, +7 new in `scoring/scores.test.ts`)
- [x] `pnpm --filter inspect e2e` (54 pass)
- [x] Visual verification via playwright screenshots against the categorical demo log


dragonstyle:
This looks good - could you add a screenshot to illustrate an example of grouped metrics in both contexts?

PR #261, opened 2026-05-28
Expose useScrollDirection from the inspect-log-viewer package

rasmusfaber:
## Summary

It would be very useful to be able to hook into `useScrollDirection` from outside.


epatey (approved):
@rasmusfaber, I'm not a fan of the re-exporting of _implementation detail_ functions used by the viewer. At the same time, we want to enable your use case, so I'm approving this PR. If this `index.ts` barrel file grows consistently over time exporting the repo's innards, we may need to rethink our approach.

I've added a disclaimer/contextual comment to `index.ts` that does the exports for embedding making it clear what the expectations are.

PR #267, opened 2026-05-31
Transcript: migrate to VirtualList + shared GeneratingIndicator

dragonstyle:
## Summary

Builds on PR #266 (Phase 1 chat migration, merged). Migrates the **transcript** view from `LiveVirtualList` to `VirtualList`, and extends the shimmer `GeneratingIndicator` from chat to pending model/tool event panels.

### Transcript migration
- `TranscriptVirtualListComponent` → uses `VirtualList` with the standard prop renames (`listHandle` → `ref`, `id` → `persistenceKey`, `initialTopMostItemIndex` → `initialIndex`, `offsetTop` → `stickyHeaderOffset`, `animation` → `smoothScroll`, `disableFindRegistration` → `findScope="none"`)
- `scrollToTopOnFinish={true}` preserves the live→done scroll-to-top behavior
- `TranscriptViewNodes.tsx` + `SamplePrintView.tsx`: swap `VirtuosoHandle` → `VirtualListHandle`
- `apps/inspect/src/app/types.ts` + `state/appSlice.ts`: swap `StateSnapshot` → `VirtualListStateSnapshot`

### GeneratingIndicator in transcript
- Replaces \`PulsingDots\` inside pending \`ModelEvent\` and \`ToolEvent\` panels with the shimmer bar
- Filters the empty assistant placeholder out of the model event's chat summary so only the indicator shows while pending
- Moves \`GeneratingIndicator\` + \`livePlaceholder\` from \`chat/\` to \`indicators/\` since both chat and transcript consume them; widens \`livePlaceholder\` to accept \`ChatMessage\` so transcript code can use it directly

### Not touched (intentional)
\`PulsingDots\` remains for: \`OutlineRow\` (small per-row sidebar indicator), \`sampleStatus\` (sample-list status cell), \`VirtualList\` default Footer (generic fallback). Different visual contexts where the full-width shimmer would be wrong.

## Remaining work (separate PRs)
- Milestone 3: Scout consumers (\`ScannerResultsList\`, \`ScannerSidebar\`)
- Milestone 4: Cleanup (delete \`LiveVirtualList\`, drop \`react-virtuoso\` from packages)

## Test plan
- [x] \`pnpm check\` clean
- [x] Manual: transcript view loads and scrolls correctly
- [x] Manual: pending model event shows shimmer bar (no empty assistant placeholder, no PulsingDots)
- [x] Manual: pending tool event shows shimmer bar
- [x] Manual: OutlineRow + sampleStatus still use compact PulsingDots


packages/inspect-components/src/transcript/ModelEventView.tsx
```diff
@@ -115,13 +114,14 @@ export const ModelEventView: FC<ModelEventViewProps> = ({
...
-  const summaryMessages = useMemo(
-    () =>
-      showAllMessages
-        ? [...event.input, ...(outputMessages || [])]
-        : [...userMessages, ...(outputMessages || [])],
-    [showAllMessages, event.input, outputMessages, userMessages]
-  );
+  const summaryMessages = useMemo(() => {
+    const outputs = (outputMessages || []).filter(
+      (m) => !isLivePlaceholderMessage(m)
```
dragonstyle: Is this meant to run only while `event.pending` is true? `isLivePlaceholderMessage()` treats assistant messages whose array content contains only `tool_use` blocks as having no visible content, so this unconditional filter drops completed model outputs that are pure server/tool-use content from the summary before `ChatView` can render them. I think this needs to be gated on `event.pending` or use a narrower predicate for the synthetic empty placeholder.

dragonstyle: Good catch — fixed in 2e2c5152. The filter now only runs while `event.pending` is true, so completed `tool_use`-only outputs survive.

PR #274, opened 2026-06-02
Improve Task/Folder/Sample list initial load status

### LINKED ISSUE #269: Improve Task / Folder / Sample List Initial Load Status in Inspect App
Currently anytime there are no rows in the Task / Folder / Sample list a small 'No Rows to Show' appears. This makes sense to show when everything has loaded with no errors and there are no rows. But we currently show this:

- while the samples or tasks are loading
- if an error occurs from the server

We should improve this to:

- Showing 'Loading...' (perhaps with pulsing ... like we use when showing message generating activity) when we are loading data. An activity indicator (the cylon bar), should be animated on screen near the top of the grid when loading.

- Showing the error when a server error occurs. Since the error is fatal we should likely replace the grid with an error display which provides the user with the full error and stack trace using one of our error displays.

stevenpollack:
## Summary

- Replace AG Grid's default "No Rows to Show" overlay with a custom `GridLoadingOverlay` (pulsing dots + visible "Loading..." text) during data fetch
- Show `ErrorPanel` with full message + stack trace when the server returns an error (replaces the grid area; navbar stays for navigation)
- Clear stale errors automatically when a new load begins (`setLoading(true)` resets the error slot)

### New components
- **`GridLoadingOverlay`** (`packages/react/src/components/`) — thin wrapper around `PulsingDots` for AG Grid's `loadingOverlayComponent` prop
- **`PulsingDots` `showText` prop** — renders text visibly alongside dots (backwards-compatible, defaults to screen-reader only)

### App wiring (`apps/inspect`)
- **`LogListGrid`** + **`SamplesGrid`** — pass `loadingOverlayComponent={GridLoadingOverlay}` to `<AgGridReact>`
- **`LogsPanel`** + **`SamplesPanel`** + **`SamplesTab`** — read `state.app.status.error` from store; conditionally render `<ErrorPanel>` in place of grid+footer when error is truthy
- **`appSlice.ts`** — `setLoading(true)` now clears any previous error so stale errors don't persist across retries

### Testing & stories
- Unit tests for `PulsingDots` and `ErrorPanel` (`packages/react`)
- Storybook stories for `PulsingDots`, `ErrorPanel`, `NoContentsPanel`, `GridLoadingOverlay`
- Bootstrap CSS loaded in Storybook preview so components render correctly
- Shared test fixtures: `__testing__/icons.ts`, `.storybook/decorators.tsx`
- Vitest config + setup file for `packages/react` (jsdom, jest-dom matchers, cleanup)
- MSW-based Playwright e2e test (`error-state.spec.ts`) verifying error display and recovery

## Known limitations

The error state uses a single global slot (`state.app.status.error`). This means:
- A failure in one operation (e.g., `loadSpecificLog`) can show in an unrelated view
- Concurrent loads can race on the error slot (later `setLoading(true)` clears earlier errors)
- Grid unmount on transient errors loses scroll position (filter/sort/column state is preserved)

These are pre-existing architectural constraints in the Zustand store, not introduced by this PR. See inline review comments for details.

Fixes: #269


apps/inspect/src/app/log-list/LogsPanel.tsx
```diff
@@ -425,32 +426,39 @@ export const LogsPanel: FC<LogsPanelProps> = ({
...
-            progress.total !== progress.complete ? (
-              <ProgressBar
-                min={0}
-                max={progress.total}
-                value={progress.complete}
-                width="100px"
-              />
-            ) : undefined
-          }
+      {error ? (
```
stevenpollack, line 429: **[Review]** ErrorPanel displays unrelated errors from the global state — a `loadSpecificLog` failure (App.tsx) shows here even when log listing succeeded.

**Failure scenario:** User views log list. Background `selectedLogFile` change triggers `loadSpecificLog` in App.tsx which fails. `state.app.status.error` is set. LogsPanel shows ErrorPanel and hides the working log list grid.

_Architectural: the single `state.app.status.error` slot is not scoped per-view._

apps/inspect/src/state/appSlice.ts
```diff
@@ -124,7 +124,11 @@ export const createAppSlice = (
             state.app.status.loading + (loading ? 1 : -1),
             0
           );
-          state.app.status.error = error;
+          if (loading) {
```
stevenpollack: **[Review]** `setLoading(true)` unconditionally clears error — concurrent loads race on the single error slot, causing errors to flash and disappear before the user can read them.

**Failure scenario:** `syncLogs` fails (sets error). Before the user reads it, replicationService starts a new sync (`setLoading(true)`), wiping the error. Grid flickers: shows ErrorPanel briefly, then back to loading overlay.

apps/inspect/src/app/samples-panel/SamplesPanel.tsx
```diff
@@ -462,48 +463,57 @@ export const SamplesPanel: FC = () => {
...
-          onRowOpen={handleRowOpen}
-          initialState={initialState}
-          onStateUpdated={setGridState}
-          onFilterChanged={updateDisplayedFromApi}
-          onFirstDataRendered={handleFirstDataRendered}
-          loading={isEmptyAndLoading}
-          autoSizeStrategy={{
-            type: "fitGridWidth",
-          }}
+      {error ? (
```
stevenpollack, line 465: **[Review]** Grid unmount on transient error loses scroll position — AG Grid's persisted `GridState` (via `onStateUpdated`) does not include scroll offset.

**Failure scenario:** User scrolls deep into a 1000-row sample grid. A transient network error during background sync shows ErrorPanel (grid unmounts). Error clears on next poll. Grid remounts at scroll-top — user loses their place.

_Filter/sort/column state IS preserved (stored in Zustand), but scroll position is not._

apps/inspect/src/app/log-view/tabs/SamplesTab.tsx
```diff
@@ -503,6 +504,15 @@ export const SamplesTab: FC<SamplesTabProps> = ({
     [allColumns, columnFilteringAllowed]
   );
 
+  if (appError) {
```
stevenpollack, line 509: **[Review]** Same global error slot issue — SamplesTab shows error for failures unrelated to this tab's data (e.g., a directory-level `syncLogs` failure).

**Failure scenario:** User is viewing per-log samples. A concurrent `syncLogs` for the directory listing fails. SamplesTab reads the global error and shows ErrorPanel even though its own sample data loaded fine.

stevenpollack:
**[Review — pre-existing issue exposed by this PR]** `apps/inspect/src/state/logsSlice.ts:251`

`initLogDir` calls `setLoading(false, error)` without a matching `setLoading(true)` — causes double-decrement of the loading counter on subsequent syncs (when `useProgress=true`).

**Failure scenario:** syncLogs increments counter (line 276), initLogDir fails and decrements here (line 251), then syncLogs hits the `!logDir` early-return and decrements again (line 317). Counter drops to -1 (clamped to 0), so the next real load's indicator disappears early.

_This pre-dates the PR but is now user-visible since errors are displayed._

apps/inspect/src/state/appSlice.ts
```diff
@@ -124,7 +124,11 @@ export const createAppSlice = (
...
+            state.app.status.error = undefined;
+          } else {
+            state.app.status.error = error;
+          }
         }),
```
stevenpollack: @claude why not use a ternary here?

packages/react/src/components/ErrorPanel.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { describe, expect, it } from "vitest";
+
+import { ComponentIconProvider } from "./ComponentIconContext";
+import { ErrorPanel } from "./ErrorPanel";
+import { testIcons } from "./__testing__/icons";
+
+const renderWithIcons = (ui: React.ReactElement) =>
+  render(
+    <ComponentIconProvider icons={testIcons}>{ui}</ComponentIconProvider>
+  );
```
stevenpollack: @claude why aren't you using the decorator in .storybook/decorators.tsx ?

packages/react/src/components/PulsingDots.module.css
```diff
@@ -56,6 +56,12 @@
   height: 12px;
 }
 
+.label {
+  font-size: 13px;
+  color: var(--bs-secondary-color, #6c757d);
```
stevenpollack: @claude double check that other CSS modules define fallbacks. If they don't, remove this.

packages/react/src/components/PulsingDots.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+describe("PulsingDots", () => {
+  it("renders with role='status' for accessibility", () => {
+    const { getByRole } = render(<PulsingDots />);
+    expect(getByRole("status")).toBeInTheDocument();
+  });
+
+  it("renders 3 dots by default", () => {
+    const { container } = render(<PulsingDots />);
+    const dotsContainer = container.querySelector("[class*='dotsContainer']");
```
stevenpollack: @claude querying on class attributes feels brittle. Suggest an alternative.

packages/react/src/components/PulsingDots.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  it("renders visible text when showText is true", () => {
+    const { getByRole } = render(
+      <PulsingDots text="Loading..." showText={true} />
+    );
+    const status = getByRole("status");
+    const label = status.querySelector(".label");
+    expect(label).toBeInTheDocument();
+    expect(label).toHaveTextContent("Loading...");
```
stevenpollack: @claude this is convoluted, you should use a more succinct query method and test directly for the text.

packages/react/src/components/PulsingDots.tsx
```diff
@@ -43,7 +47,7 @@ export const PulsingDots: FC<PulsingDotsProps> = ({
           />
         ))}
       </div>
-      <span className={styles.visuallyHidden}>{text}</span>
+      {!showText && <span className={styles.visuallyHidden}>{text}</span>}
```
stevenpollack: @claude this is not the ARIA/WCAG "best practices" for text that is for accessibility purposes. Justify this.

packages/react/src/components/NoContentsPanel.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+};
+
+export default meta;
+type Story = StoryObj<typeof NoContentsPanel>;
+
+export const Default: Story = {
+  args: {
+    text: "No samples",
+  },
+};
```
stevenpollack: @claude default stories shouldn't have args, correct?

stevenpollack:
**ActivityBar audit (14dbcfe)**

Traced `ActivityBar` wiring across all three panels that show a grid loading overlay:

| Panel | Before | After |
|---|---|---|
| `LogsPanel` | ✅ via `ApplicationNavbar` (`showActivity="log"`) | unchanged |
| `SamplesPanel` | ⚠️ double — one inside `ApplicationNavbar`, one freestanding below | removed the duplicate freestanding one |
| `SamplesTab` | ❌ none (no `ApplicationNavbar` in this tab) | added `<ActivityBar animating={!!loading} />` at top of `Fragment` |

The `SamplesPanel` duplication was introduced by this PR — the manual `<ActivityBar>` was added without noticing `ApplicationNavbar` already renders one internally.

dragonstyle:
A couple of very quick hits:

- The Samples tab layout is currently broken because the ActivityBar is removed but its grid slot was not.
- Use `pnpm format` to ensure code is properly formatted (use `pnpm check` to run listing, formatting, and tests)

apps/inspect/src/app/log-list/LogsPanel.tsx
dragonstyle: Agree - I think keeping data and error in state is a bad pattern we use in this app. We should be handling data and errors right in the panel when the data is loaded (e.g. using a hook to load which exposes AsyncData). This is something that is in the works / on the radar so this is something we could live it.

dragonstyle:
There is a root problem which is that background syncing really shouldn't be sharing the global error slot - it will cause a lot of user disruption (as I continued reading the comments). I think we may need to take a different approach with syncing errors - something like an error adornment on the gear icon + error message there, or a toast, for example.

dragonstyle:
Closing as superseded by #304 (same branch history plus the follow-ups from the June review), which I've closed with the reasoning there.

PR #278, opened 2026-06-03
Move storybook from react/ to tooling/ for repo-wide usage

stevenpollack:
Storybook can now be started from the repo root via
```
pnpm storybook
```
and the storybook build is cached by turbo:
```
pnpm storybook:build
```

## Story organization follows repo structure
image

## autodocs are default
image

## Full-app MSW interaction stories

Beyond leaf-component stories, the inspect viewer now has **interaction stories**
that render the real `<App/>` (hash router + Zustand store + bootstrap effects)
and drive whole application states entirely through [MSW](https://mswjs.io/)-mocked
network responses — no Python backend required:

- `App/FullApp` → `LogListing`, `CompletedEvalSynthetic`, `CompletedEvalRealData`,
  `RunningEval` (live `/pending-samples` polling), `ErrorState`.

Mock handlers + typed data factories live in `apps/inspect/src/mocks/` and are
shared with the Playwright e2e suite (re-exported from `e2e/fixtures/`).

### Gotchas (documented in `tooling/storybook/README.md` + `AGENTS.md`)

- **Per-story `parameters.msw.handlers` REPLACE the global handler array** rather
  than concatenating. A story that sets its own handlers otherwise loses every
  boot endpoint and 404s on bootstrap. Each story wraps its overrides with
  `withDefaults([...])`.
- **A passing `storybook:build` does not mean the stories work** — it only proves
  they compile. Verify in a real browser via
  `pnpm --filter @meridianlabs/log-viewer story-smoke` (headless Chromium loads
  each story's iframe and asserts rendered content + no unexpected errors).
- Known limitation: overview-route stories log a benign background
  `No database available for replication` (an app-level `syncLogs` race); views
  render correctly regardless.


turbo.json
```diff
@@ -22,8 +31,19 @@
...
+    "storybook": {
+      "cache": false,
+      "persistent": true
+    },
     "storybook:build": {
-      "outputs": ["storybook-static/**"]
+      "inputs": [
+        ".storybook/**",
+        "package.json",
+        "../../**/*.stories.{ts,tsx}"
```
stevenpollack: **Stale Storybook builds: cache `inputs` only hash story files, not what they import.** The `storybook:build` inputs are `.storybook/**`, `package.json`, and `../../**/*.stories.{ts,tsx}`. But `App.stories.tsx` (and others) import non-story sources that drive the output — `src/mocks/factories.ts`, `mocks/handlers.ts`, `mocks/story-decorator.tsx`, `mocks/fixtures/completed-eval.json`, and the component sources themselves. None match `*.stories.{ts,tsx}`. Editing a fixture or handler → Turbo reports a cache hit and ships the **old** Storybook — exactly the "passing build ≠ working" trap your README warns about, re-introduced at the cache layer. Consider modeling this as a task dependency (`dependsOn: ["^build"]` + the consuming packages as deps) rather than a story-only glob, or broaden the inputs to the imported source dirs.

tooling/storybook/.storybook/main.ts
```diff
@@ -0,0 +1,35 @@
+import type { StorybookConfig } from "@storybook/react-vite";
+
+const config: StorybookConfig = {
+  stories: [
+    "../../../**/*.stories.@(ts|tsx)",
```
stevenpollack: **Repo-root `../../../**` glob can pull in `*.stories.tsx` from `node_modules`.** This wildcard is anchored at the repo root and is not guaranteed to exclude `node_modules` (there are third-party `*.stories.tsx` files under it). That can error at index time or pollute the sidebar. Prefer scoping to the packages that actually carry stories (e.g. `../../../apps/*/src/**/*.stories.@(ts|tsx)` + `../../../packages/*/src/**/*.stories.@(ts|tsx)`) — the "intentional registration" the design seems to want. Verify in a browser, not just `storybook:build`.

apps/inspect/scripts/story-smoke.mjs
```diff
@@ -0,0 +1,82 @@
...
+  page.on("pageerror", (e) => pageErrors.push(e.message));
+  page.on("response", (r) => {
+    const u = r.url();
+    if (u.includes("/api/") && r.status() >= 400 && !EXPECTED_404.some((p) => u.includes(p))) {
+      apiFail.push(`${r.status()} ${u.replace(/\?.*/, "")}`);
+    }
+  });
+
+  await page.goto(`${BASE}/iframe.html?id=${id}&viewMode=story`, { waitUntil: "domcontentloaded" });
+  await page.waitForTimeout(7000);
```
stevenpollack: **Fixed 7s sleep per story, run sequentially → ~35s floor that grows with every story.** Each iteration waits a hard `waitForTimeout(7000)` regardless of how fast the story renders. The script already knows the expected text per story (`EXPECTATIONS[id].expect`) — replace the fixed sleep with `page.waitForFunction`/`expect(locator).toContainText(...)` polling capped by a timeout, and optionally run the per-story contexts concurrently. Typically cuts the run to a few seconds.

apps/inspect/src/mocks/factories.ts
```diff
@@ -0,0 +1,192 @@
...
+    .reverse()
+    .find((m) => m.role === "assistant");
+
+  return {
+    id: overrides.id,
+    epoch: overrides.epoch ?? 1,
+    input: overrides.messages
+      .filter((m) => m.role === "user")
+      .map((m) => m.content)
+      .join("\n") as string,
```
stevenpollack: **`as string` masks corruption of structured message content.** `createEvalSample` does `messages.filter(user).map(m => m.content).join("\n") as string`, assuming every user `content` is a string. A user message with array/structured content (valid per `ChatMessage`) yields `"[object Object]"` in `input` (and the same pattern at the `output` fallback returns the literal `"Response"` for array assistant content). No current story triggers it, but it is a latent data-builder bug — handle non-string content explicitly instead of casting.

apps/inspect/src/app/App.stories.tsx
```diff
@@ -0,0 +1,313 @@
...
+// RunningEval: polling /pending-samples until completion
+// ---------------------------------------------------------------------------
+
+export const RunningEval: Story = {
+  parameters: {
+    initialRoute: `/logs/${ENCODED_LOG}`,
+    msw: {
+      handlers: (() => {
+        // Closure counter resets each render since the story object is re-evaluated
+        let callCount = 0;
```
stevenpollack, line 257: **`RunningEval` only plays on the first visit in a session.** `callCount` lives in an IIFE that runs once at module import, not per render, so revisiting this story (after navigating away) starts already-completed — the running→completed transition never replays. The smoke test masks this (fresh browser context per run). I corrected the misleading "resets each render" comment; a full fix needs to reset the counter per mount (e.g. an MSW loader or resetting in the first boot handler).

dragonstyle:
Thanks for putting this together, and for the write-up on the MSW pitfalls.

I'm going to close this one. It's been sitting since mid-June and the ground under it has moved a lot: the viewer's data layer went to TanStack/react-query (#389), backend resolution was refactored into `app_config/` (#482), and a couple of the components the stories target have moved or been removed. The full-app stories in particular would need to be rebuilt rather than rebased, and we don't currently have anyone writing Storybook stories on the team, so I'd rather not ask you to do that work speculatively.

If we do decide to invest in a repo-wide Storybook, the `tooling/storybook` host here is the shape we'd want, and I'd be glad to pick it back up, or take a fresh, smaller PR with just the host and root scripts.

PR #304, opened 2026-06-09
Improve Task/Folder/Sample list initial load status

stevenpollack:
## Summary

This PR improves the initial load UX for the Task, Folder, and Sample list views, and cleans up several related issues found during review.

### Loading indicator
- Adds `GridLoadingOverlay` to `@tsmono/react` — a thin wrapper that shows a `PulsingEllipsis` inside AG Grid's loading overlay slot
- Wires `GridLoadingOverlay` into `LogListGrid` and `SamplesGrid` so the grid shows an animated "Loading..." indicator while the initial fetch is in flight, rather than a blank or spinning cell
- Extracts `PulsingEllipsis` as a new primitive in `@tsmono/react` — three stagger-fading dots after a label, with `role="status"` and `prefers-reduced-motion` support. Both `GridLoadingOverlay` and `GeneratingIndicator` now share this primitive

### Error state
- On initial load failure, an `ErrorPanel` replaces the grid (in `LogListGrid` and `SamplesPanel`)
- Background sync errors (polling) are **not** routed to the global error slot. Instead, a new `syncError` field is added to `AppStatus`; `refreshCallback` in `useClientEvents` catches polling errors and writes them there
- The options button icon is changed from info → gear (`bi-gear`), and a red dot badge appears on it when `syncError` is set. Tooltip shows the error message

### ActivityBar
- `ApplicationNavbar` already renders `ActivityBar` internally. Removes a duplicate `ActivityBar` that had been added directly inside `SamplesPanel`
- Adds the missing `ActivityBar` coverage to `SamplesTab` (it was absent)

### Storybook & tests
- Adds Storybook stories for `GridLoadingOverlay`, `PulsingDots`, `PulsingEllipsis`, `ErrorPanel`, `NoContentsPanel`
- Adds unit tests for `GridLoadingOverlay`, `PulsingDots`, `PulsingEllipsis`, `ErrorPanel`
- Adds a `__testing__/providers.tsx` and `__testing__/icons.ts` in `@tsmono/react` for shared test setup
- Adds Bootstrap CSS import to Storybook preview so components render with correct tokens
- Adds e2e tests (MSW) for the 500 error state and retry recovery in the log list
- Adds a unit test for `clientEventsService` verifying polling errors don't propagate out of the service

## Test suite
[video.webm](https://github.com/user-attachments/assets/796dbf04-3fac-46d9-acb5-cb7cc3162eee)
[video.webm](https://github.com/user-attachments/assets/3d30e877-861a-4f36-aa59-b4c1cc8aa63b)
[video.webm](https://github.com/user-attachments/assets/32f55cdb-571b-48e9-aed5-556af6c251ea)
[video.webm](https://github.com/user-attachments/assets/d373d1f2-8d04-405a-9ff9-6d339dc4ff14)


apps/inspect/src/state/clientEventsService.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        throw boom;
+      });
+      clientEventsService.setSyncErrorCallback(syncError);
+
+      // Trigger a refresh by accessing the private method via a public path:
+      // startPolling fires the refresh; we can also call it indirectly by
+      // invoking the internal path through a mock API that returns a refresh event.
+      // Simpler: reach refreshLogFiles via the stored callback indirectly by
+      // calling the public method with a mock API.
+      const mockApi = {
```
stevenpollack: @claude these comments violate guidelines.

apps/inspect/src/state/clientEvents.ts
```diff
@@ -9,20 +9,26 @@ const log = createLogger("Client-Events");
...
   );
 
   useEffect(() => {
     clientEventsService.setRefreshCallback(refreshCallback);
   }, [refreshCallback]);
 
+  useEffect(() => {
+    clientEventsService.setSyncErrorCallback(setSyncError);
+  }, [setSyncError]);
+
```
stevenpollack: @claude does this actually need `useEffect` or can this be set in the component body? Zustand state variables would trigger a re-render upon update...

apps/inspect/src/state/appSlice.ts
```diff
@@ -125,7 +126,12 @@ export const createAppSlice = (
             state.app.status.loading + (loading ? 1 : -1),
             0
           );
-          state.app.status.error = error;
+          state.app.status.error = loading ? undefined : error;
```
stevenpollack, line 130: @claude do we still want to be changing this if we've now created a specific `syncError` state variable?

apps/inspect/src/app/samples-panel/SamplesPanel.tsx
```diff
@@ -461,49 +462,57 @@ export const SamplesPanel: FC = () => {
         scoresHeading="Scores"
       />
 
-      <ActivityBar animating={!!loading} />
```
stevenpollack, line 464: @claude have you introduced a regression by removing this ActivityBar?

stevenpollack:
**On sync error dismissal:** The app has a `MessageBand` component (inline dismissible banner) but no floating toast system. We opted not to use it for `syncError` because the dismissed state is keyed by ID in Zustand — once dismissed, a later sync error with the same ID would be silently swallowed.

Instead, `syncError` is self-healing: `refreshCallback` clears it at the start of every sync cycle, so the gear dot and popover error row disappear automatically on the next successful sync. No user dismissal needed.

The error is now surfaced in two places:
1. Red dot on the gear icon (visual indicator)
2. "Sync Error" + message row at the top of the gear popover (readable detail on click)

dragonstyle:
Thanks for this, and for turning the June review feedback around so quickly — the gear-icon adornment was the right shape for background-sync errors.

Closing it now because the ground moved underneath it. In July the log and sample lists moved from AG Grid to our own TanStack-based DataGrid with a react-query data flow (#389), and AG Grid was removed entirely shortly after (#408). The `clientEvents` service this hooks into no longer exists, and the branch conflicts in 13 files including two the refactor deleted. The loading indicator and the error panel that #269 asked for came along with that refactor, so the only piece still missing is the background-sync error adornment. I'll narrow #269 to that rather than ask you to rebuild the rest on a grid you haven't seen.

If you'd like to pick that adornment up against current main, it's a small change now (`ViewerOptionsButton` still uses the info icon, and `LogsPanel` still folds `sync.error` into the full-panel error). Either way, thanks for the work here.

PR #307, opened 2026-06-09
remove bootstrap

Aprillion:
* TBD with Charles (this is slightly more radical than original ask, but the diff looks reasonable to me just a tiny bit large 😵‍💫, but after some a "few" rounds, I can no longer spot any differences from main branch in the browser for both inspect_ai and scout (and update to hawk needs to be merged after release of those two))


dragonstyle:
I don't mind this direction. If we overlap time wise would love to have you just give me high level thinking to make sure I'm following properly. 

Fable caught the following nits...

1. EmptyPanel lost all its layout — .container is used but no longer defined anywhere.
packages/react/src/components/EmptyPanel.tsx:10 renders <div className={"container"}>, but components.css deliberately kept only .container-fluid. Pre-PR, Bootstrap's .container gave it width/auto-margins/padding; now it's a bare div. This affects the "No Sample Selected" and loading states (LogView.tsx, SampleScoresGrid.tsx). There's a dead EmptyPanel.css next to it that was never imported (pre-existing). Fix: give EmptyPanel a proper CSS module (and delete the dead file), or add .container to components.css.

2. VS Code dark mode: tool-error styling lost its IDE-palette override.
Old base.css (lines 350–351 at the merge base) mapped --inspect-tool-error-color/-bg to var(--vscode-errorForeground) / var(--vscode-inputValidation-errorBackground) inside the VS Code block. The new body[class^="vscode-"] block in tokens.css omits both, so they fall through to rgba(--inspect-danger-rgb, 0.06) — near-invisible against dark VS Code themes instead of the proper #3a1d1d-style background. One-line fix: restore both overrides in the vscode block.

3. Scout SearchPanel (SearchPanel.module.css): six selectors go from inherited text color to primary-emphasis blue (#052c65 light / #6ea8fe dark). This changes, but might be improved- worth inspecting.

4. FindBand fond-size → font-size typo fix (buttons grow to 1.1rem); ToolInput text-size → font-size (code shrinks to 0.8rem); ColumnFilterEditor gains monospace; MessageBand error bands gain danger colors; ScanPanelTitle gains muted color. This changes, but might be improved- worth inspecting.

5. Shared packages silently depend on theme CSS they neither import nor declare. packages/react has zero references to @tsmono/theme (verified), yet ToolButton.module.css:1 styles via :global(.btn), TabSet/NavPills render nav/nav-tabs/tab-content classes, and useBreadcrumbTruncation.ts:40 even measures an element assuming .breadcrumb padding is loaded. inspect-components similarly relies on spinner-border, text-size-*, and bootstrap-icons. It all works today only because the apps load theme CSS top-down. This is the single biggest blocker to shipping these packages: either import the needed @tsmono/theme/* CSS at each package entry point with @tsmono/theme as a peerDependency, or document a hard host contract. I'd prioritize this right after the P0s — it's also the path that eventually shrinks components.css global classnames into modules.

6. The token system isn't used by its own extracted CSS in places. components.css hardcodes #86b7fe and rgba(13,110,253,.25) for .form-control:focus/.form-select:focus (lines ~93, ~213, plus two more focus-ring spots) even though --inspect-focus-border-color and --inspect-focus-ring-color exist precisely for this. Faithful to Bootstrap (which also hardcodes them), but it blocks VS Code/host retheming of focus rings. Cheap to fix now while the file is fresh.

7. Empty :root[data-theme="dark"] {} block at the end of tokens.css; empty a:hover {} in reboot.css (the real rule lives in base.css — at least add a pointer comment); two orphaned zero-byte files MorePopover.css/MorePopOver.css; duplicate readable-variant rule in both tokens.css and base.css; a comment in the contract test noting it skips design//docs//.md and that the lib skip-dir matches the vite outDir.

Aprillion:
OK, let's find a time to sync (TBD on Slack).

Some of the other-Fable's points now fixed by the original-Fable (the decisions and the code LGTM):
```
  #: 1
  Point: EmptyPanel lost .container
  Status: Fixed — CSS module replicating Bootstrap container sizing; dead EmptyPanel.css deleted
  ────────────────────────────────────────
  #: 2
  Point: VS Code tool-error palette
  Status: Fixed (half), pushed back (half) — restored the --inspect-tool-error-bg override; the color
    needed nothing, it already resolves to --vscode-errorForeground transitively via
    --inspect-danger-foreground
  ────────────────────────────────────────
  #: 3
  Point: SearchPanel blue selectors
  Status: Pushed back on the branch's change — --bs-primary-color never existed, so those five always
    rendered inherited; I restored color: inherit (with comment) instead of the blue. If a designer
    wants blue/muted there, that's a deliberate restyle for a follow-up
  ────────────────────────────────────────
  #: 4
  Point: Typo-fix visual changes
  Status: Kept all five deliberately (FindBand, ToolInput, ColumnFilterEditor, MessageBand,
    ScanPanelTitle) — each was an invalid var and the correction matches obvious intent (e.g. unstyled
    error bands were plainly a bug). gpt-5.5 concurred: "none looks like a mistaken regression"
  ────────────────────────────────────────
  #: 5
  Point: Packages' undeclared theme dependency
  Status: Documented, implementation deferred — all @tsmono/* packages are private: true, so this
    blocks nothing today; the host contract + publishing precondition is now spelled out in the
    components.css header. Entry-point imports + peerDependency is the right move when publishing  —
    needs discussion only if that's planned
  ────────────────────────────────────────
  #: 6
  Point: Hardcoded focus rings
  Status: Fixed — all four spots now use --inspect-focus-border-color/--inspect-focus-ring-color (the
    per-variant --inspect-btn-focus-shadow-rgb triples stay, faithful Bootstrap structure)
  ────────────────────────────────────────
  #: 7
  Point: Cleanups
  Status: All done — empty dark block, duplicate readable rule, a:hover stub → pointer comment,
    zero-byte MorePopOver.css, contract-test scope comment

  gpt-5.5 adversarial re-review: 8/8 PASS (the 7 points + the e2e fixture fix), including spec-level
  confirmation that color: inherit is exactly equivalent to an invalid custom-property reference for
  color, and that the */api/logs msw pattern can't shadow /api/logs/{file}.

  Bonus find while verifying: the e2e suite's flakiness was real and got worse mid-session — the
  default fixtures never mocked /api/logs, so every test's boot raced a proxy 502 against setLoading
  error-clearing. Fixed with a default handler (298a9f24); the suite is deterministic now (62/62 twice
  consecutively, previously-failing spec 3/3).
```

Aprillion:
few high-level points from the call:
- we want a targeted public interface how downstream clients (hawk, ...) can use css variables for their theming, so we shouldn't spread the --inspect- variables through multiple large files, but perhaps single theme.css that would drive the whole experience and can be overwritten
- next step to keep in mind would be customizable icons, how can consumers choose different icons? css variables might be limited (e.g. currentColor in SVG), while fine-grained props on component could be overkill before it's actually needed, so just keep in mind
- in the future, we might expose reusable components, not just whole (embedded) apps

...I wanted to give it another go, but I think my Fable session started to overengineer stuff after I used "design" and "architecture" in the same prompt 🙈 so I am parking the draft commits here and will continue to pick the good parts another day

Aprillion:
~~ok, the change sounds good overall, but the commits were terrible => now rebased for easier review 🤞 (but I will re-check manually one more time later today before marking as ready for external review)~~

edit: nope, the code from Fable was pure slop and it just sounded good after summarized by all the "adversarial" agentic reviewers (both opus 4.8 and gpt 5.5 were glazing all over it, but when I looked at it again next day with the actual goal to have nice theming interface in mind, that part of the code doesn't make sense ... at all? aaaargh 😱 ... working on it more)

packages/theme/src/theme.css
Aprillion, line 1: I think I am happy with the file structure now, `theme.css` for lightweight public interface that should remain relatively stable (we can add to it what we want to support as needed in downstream consumers), while `tokens.css` and other shared css should keep all the internal details needed for complete re-skins like the vscode webview and remaining component libraries (and everything else should be moved to *.module.css files):

Image

packages/react/src/components/PopOver.tsx
Aprillion, line 1: this is a big diff, but it fixes a FOUC (popovers appearing in wrong position for fraction of a second) that got triggered by bootstrap removal but claude convinced me that the root cause & fix belongs here and the new code looks reasonable to me (as well as other agents), so I kept it in this PR

apps/inspect/src/app/samples/header-v2/ScorePanel.module.css
```diff
@@ -55,25 +55,28 @@
...
 }
 
-:global(.btn).sortButton {
-  background: transparent;
-  border: none;
-  border-radius: var(--bs-border-radius);
-  color: var(--bs-secondary);
+/* Re-skin the Button vars (this module's CSS loads after Button's). */
+.sortButton {
+  --_bg: transparent;
```
Aprillion, line 60: I am not sure how claude convinced me that this is the best solution for replacement of `--bs-*` variables and `.btn` classes which shouldn't pollute reusable global `--inspect-*` but stay internal between our components (Opus 4.8 first, but when gpt-5.5 reviewer agreed with it so I gave up)

looking at this today, `--_bg` seems stupid again, but I don't have better idea at the moment, flagging for next week..

@dragonstyle - let me know if you have a suggestion, I am probably missing something obvious (after spending too much time on this branch 😵‍💫)

Aprillion (commented):
conceptual changes in https://github.com/meridianlabs-ai/ts-mono/pull/307/changes/57bfc18a57b949992e984a22bc493aa226a88584, the other commits should be mechanical rename and conversion from `.btn` to components

..mostly de-sloppified 🤞 but I probably still missed some stupid over-engineering somewhere, so let me know what needs more attention (if we chose to remove Bootstrap after 🙈)

Aprillion:
let's close this one for now - if we decide to remove bootstrap and reopen this PR alter, there will be a lot of rebasing merge conflicts to do (...but after https://github.com/meridianlabs-ai/ts-mono/pull/354 and other big features)

PR #309, opened 2026-06-09
Integrate /app-config

brandly:
Integrating the new `/api-config` endpoint https://github.com/UKGovernmentBEIS/inspect_ai/pull/4142

Currently it only holds Inspect and Scout versions, but soon it will replace `/log-dir`. For this reason, we'll treat it as something we need to fetch upfront and gate the entire application on waiting for this data to load.

We'll show the versions in a popover in the top-right corner:
Screenshot 2026-06-09 at 10 46 12 AM

Previously we had a "version" in this popover that would come from git tags at build time. However this stopped working when the viewer was moved to `ts-mono` since we don't publish any release tags in this repo. It's been removed!

We add `"@tanstack/react-query"` to our dependencies, just like Scout's viewer. For now, we'll only use it for this new requests, but we plan to make more use of it in the future.


dragonstyle (approved):
👍

PR #316, opened 2026-06-10
sample loading perf improvements

Aprillion:
Hydrate log listing via batched server endpoints (125s -> 23s for 1418 logs)

- get_log_summaries delegates all formats to the API's batched endpoint (view server and VS Code read .eval headers server-side; static deployments resolve from the manifest); the per-file client-side zip path and get_log_summary are deleted as dead code, converging with the unmerged upstream fix (inspect_ai faber/remote-performance, 0064b921d)
- new LogViewAPI.get_log_details_batch backed by the view server's new /log-details endpoint (header + sample summaries, undefined for unreadable files), with per-file fallback for APIs without it (VS Code)
- replication detail queue batches 12 (was 1), preview batches 24 (was 6)
- batched GET query strings chunked at ~6KB for request-line limits
- polling: resolve attachments once per event (after pool expansion) instead of twice

- ^^ the measurements are by Claude Fable 5 itself, I just smoke-tested in browser and Cursor extension with local log and s3, didn't notice any bugs, everything feels 10x faster subjectively, and the code looks reasonable when manually reviewed
- API changes in: https://github.com/UKGovernmentBEIS/inspect_ai/pull/4210

### DIFF
--- a/apps/inspect/src/client/api/client-api.ts
+++ b/apps/inspect/src/client/api/client-api.ts
@@ -234,70 +234,18 @@ export const clientApi = (
     return undefined;
   };
 
-  const read_eval_file_log_summary = async (log_file: string) => {
-    // If the API supports this, delegate to it
-    if (api.get_log_summary) {
-      return api.get_log_summary(log_file);
-    } else {
-      // Don't re-use the eval log file since we know these are all different log files
-      const remoteLogFile = await openRemoteLogFile(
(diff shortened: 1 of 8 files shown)


Aprillion:
starting with more targeted improvements first => https://github.com/UKGovernmentBEIS/inspect_ai/pull/4211

PR #329, opened 2026-06-12
Add search panel to Inspect

brandly:
This PR takes the search panel that was added to Scout (#155), lifts the component up into a shared package, and integrates it into Inspect.

https://github.com/user-attachments/assets/66a96547-a2d2-4c1d-b8b4-60c5d0d59596

`SearchPanel` exposes an adapter interface that each app implements. Components like `SidebarHeader` and `ResizablePanel` were also lifted into a shared package. 

Introducing this search panel to Inspect raised some UX questions, since there was an existing scans panel. Now we have unified tabs in a side menu for opening each panel. The "scans" tab is only present if the given transcript has any scans.


dragonstyle:
### Code review

Found 2 issues:

1. Arrow keys hijack sample navigation while typing in the new search panel. Inspect's global ArrowLeft/ArrowRight handler guards on `document.activeElement.tagName === "INPUT" | "TEXTAREA" | "SELECT"`, but the shared SearchPanel's inputs are shadow-DOM custom elements (`vscode-textarea` / `vscode-textfield`), so `activeElement.tagName` is `"VSCODE-TEXTAREA"`, the guard fails, and left/right while composing a query navigates samples. Scout hit this same bug (flagged in #155) and fixed it in `NextPreviousNav.tsx` by walking `shadowRoot.activeElement` — that guard needs porting here now that the panel renders in Inspect.

https://github.com/meridianlabs-ai/ts-mono/blob/d9ddc906ad49042144a2f6a4328031474172033a/apps/inspect/src/app/samples/SampleDetailComponent.tsx#L130-L160

2. `scansDefaultedRef` never resets across log navigation. The auto-open-Scans-on-first-view effect is gated by a ref set once and never cleared, and the effect deps omit `dockKey`. `SampleDisplay` stays mounted across prev/next navigation, so after the first sample with scans, subsequent logs with scans and no persisted dock choice (`storedDock === undefined`) never get the auto-open default. Resetting the ref when `dockKey` changes (or keying it by `dockKey`) would fix it.

https://github.com/meridianlabs-ai/ts-mono/blob/d9ddc906ad49042144a2f6a4328031474172033a/apps/inspect/src/app/samples/SampleDisplay.tsx#L428-L440

brandly:
@dragonstyle both should be fixed now!

dragonstyle:
Looks great! Four other things I noticed:

- I'm sort of surprised by how slow llm search is. It might just be that I'm used to results streaming in and making it feel faster, but this always takes a lot a longer than I would've expected. Worth just checking to make there isn't some weird overhead that we're not expecting?
- I'm not seeing event or message cite labels once I do a search
- I'm not sure we should have the margin at the top of the activity rail before the item. Does it look weird with no top margin?
   Screenshot 2026-06-15 at 7 59 50 AM
- I frequently start typing a search term, then realize I'm on the wrong segment (llm) and toggle to grep, which clears the search text I entered. I think we shouldn't clear that text if a user hasn't searched yet.

dragonstyle:
Fixes #333

brandly:
@dragonstyle 

> Worth just checking to make there isn't some weird overhead that we're not expecting?

On a single segment transcript, I was surprised to find that we make a lot of external API calls to count tokens.

```
  ┌───────────────────────────────────────┬─────────────────────────────┬────────────────────────────────────────────┐
  │                 Phase                 │         Wall-clock          │                 What it is                 │
  ├───────────────────────────────────────┼─────────────────────────────┼────────────────────────────────────────────┤
  │ Transcript load (parquet/duckdb)      │ ~0.2s                       │ fine                                       │
  ├───────────────────────────────────────┼─────────────────────────────┼────────────────────────────────────────────┤
  │ count_tokens — 23 Anthropic API calls │ ~0.8–1.1s                   │ segmentation budgeting + template overhead │
  ├───────────────────────────────────────┼─────────────────────────────┼────────────────────────────────────────────┤
  │ generate — 1 Anthropic API call       │ ~3s (server)                │ the actual search — inherent               │
  └───────────────────────────────────────┴─────────────────────────────┴────────────────────────────────────────────┘
```

That number of requests will scale up with token counts. Potentially we could make local estimates if we don't actually need exact counts and cut some of the time down. `model.generate()` seems to be the bulk of the time though.

On larger runs, it looks like it's spending a greater portion of the total time on these token counts e.g. 9.5 seconds out a 24 second call. We still have two consecutive `model.generate()` calls in a row since we can parallelize the individual segments, but we have to wait for all of them to finish to run the LLM reducer. But we could cut down on almost 40% of the total time by improving the token counting and segmentation approach. 

EDIT: I was woefully out of date, and this token counting for Anthropic models has [improved on `main`](https://github.com/meridianlabs-ai/inspect_scout/pull/465)

> I frequently start typing a search term, then realize I'm on the wrong segment (llm) and toggle to grep, which clears the search text I entered. I think we shouldn't clear that text if a user hasn't searched yet.

Okay interesting!! We'd completely isolated the two sets of state, but this would be some kind of pre-submit state that carries over. I'll make it happen

dragonstyle (approved):
- It looks like maybe we aren't showing the event & message numbers when displaying the transcript with search results.

PR #337, opened 2026-06-17
Consolidate modal implementations

brandly:
Inspect had a legacy modal component and in some cases was using this component rather than the shared component in `packages/react/`. The shared component also had some a11y shortcomings.

This PR removes the legacy modal and migrates uses of it to the shared component. The shared component's props were extended to support these use cases. Its accessibility also improved in a number of ways including better tab handling and hitting `esc` to close the modal. 

There are some minor visual differences you can see here:

edit-metadata
edit-tags
scoring-detail


dragonstyle (approved):
🔥

PR #338, opened 2026-06-18
Integrate ActivityRail into Scout

brandly:
In https://github.com/meridianlabs-ai/ts-mono/pull/329, the `SearchPanel` was lifted out of Scout and integrated into Inspect. This raised some UX questions around the relationship between the search panel and the scans panel which resulted in the creation of the `ActivityRail`. This rail lives along the right side of the transcript and provides menu items for opening each panel.

We now want to integrate this piece of UI back into Scout for accessing the search panel and validation panel. As a part of these changes, the issue where scrolling while a panel was open would fail to collapse the header has been fixed.

I think the main thing to consider here is that "Validation" is a much longer word than search or scans, which leads the rail having a width of 62px as opposed to 44px in Inspect. Here's what it looks like on my 14" screen

https://github.com/user-attachments/assets/f3924fd2-f40a-455a-8ae9-1028a8291d6d


docs/superpowers/specs/2026-06-17-scout-activity-rail-design.md
brandly, line 1: Claude committed this spec for the changes. I'm not sure we care to keep it around

dragonstyle: I've been letting these in - I'm good committing these in the event that we end up want to dig b ack into them if we find / fix a regression or need background context for future work...

dragonstyle (approved):

PR #339, opened 2026-06-18
Fix stale "running" status in log listing after navigating back

Aprillion:
* when I tested running evals, the listing updated to finish status originally, but navigating to log details and back to the listing showed it as running again, this fix keeps showing it as finished (but let me know if I'm missing some bigger picture)


dragonstyle:
Thx for catching this! I have a slightly different fix (don't treat running evals as cacheable) that I'll merge...

PR #343, opened 2026-06-22
perf(viewer): production build config — minification, vendor splitting, content-hashed immutable assets

sjawhar:
Re-implements two closed Inspect viewer perf changes — production minification and content-hashed immutable assets — ported into `ts-mono` after the `_view/www` → submodule migration. Combined into one PR since both are `apps/inspect/vite.config.ts` build-config changes and the content-hashing builds directly on the minification/chunking setup.

## What
- Production minification + vendor chunk splitting.
- Content-hashed asset filenames, so hashed assets can be served with immutable `Cache-Control` (the matching server-side change ships in our downstream fork).

## Verification
- `pnpm --filter @meridianlabs/log-viewer build` — green (minified, content-hashed filenames emitted)

Supersedes #348 (folded in here). One of a series re-porting the closed Inspect viewer PRs into ts-mono.

### DIFF
--- a/apps/inspect/vite.config.ts
+++ b/apps/inspect/vite.config.ts
@@ -108,7 +108,7 @@ export default defineConfig(({ mode }) => {
         },
         cssCodeSplit: false,
         sourcemap: true,
-        minify: false,
+        minify: true,
       },
     };
   } else {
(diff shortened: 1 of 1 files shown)


epatey:
@sjawhar, thanks for taking a swing at this. I actually went down this path a few months ago doing exactly this turning on git LFS to avoid it blowing up the repo. The reason LFS was needed was that, with minification enabled, every script change ends up basically modifying all of the built script - blowing up the repo size.

This all worked (along with the fingerprint based immutable caching), but we had to revert it because it caused us to exceed UKAISI's github budget for LFS cost. 

So the current state is that scout does exactly what you do in this PR, but inspect continues to commit unminified built script.

PR #346, opened 2026-06-22
perf(viewer): align transcript find ordering with structured tool output

sjawhar:
Re-implements a closed Inspect viewer change, ported into `ts-mono` after the `_view/www` → submodule migration.

## What
Aligns in-transcript find ordering and match counts with how structured tool output is rendered (`transcript/eventText.ts`): an array tool result is walked into one searchable segment per rendered block in render order (dropping non-text payloads) instead of a single JSON blob. The PR documents exactly which cases are renderer-exact vs. approximate.

## Verification
- `pnpm --filter @tsmono/inspect-components test -- eventText` (incl. new tests) — green

One of a 7-PR series re-porting the closed Inspect viewer PRs into ts-mono.

### DIFF
--- a/packages/inspect-components/src/transcript/eventText.ts
+++ b/packages/inspect-components/src/transcript/eventText.ts
@@ -113,7 +113,9 @@ export const extractEventFields = (event: EventType): [string, string][] => {
         if (typeof toolEvent.result === "string") {
           fields.push(["result", toolEvent.result]);
         } else {
-          fields.push(["result", sanitizeStringify(toolEvent.result)]);
+          for (const text of extractToolResultText(toolEvent.result)) {
+            fields.push(["result", text]);
+          }
         }
       }
       // Tool error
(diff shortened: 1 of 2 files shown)


packages/inspect-components/src/transcript/eventText.ts
```diff
@@ -474,3 +476,59 @@ const extractContentText = (content: string | Array<Content>): string[] => {
...
+  if (typeof item !== "object") {
+    return [];
+  }
+  if (!("type" in item)) {
+    return [sanitizeStringify(item)];
+  }
+
+  switch (item.type) {
+    case "text":
+      return "text" in item && typeof item.text === "string" ? [item.text] : [];
```
dragonstyle, line 513: Non-blocking, and probably already inside the PR's documented "approximate" set — flagging to confirm the boundary is intentional.

When a `text` block's `.text` is itself a JSON string, `ToolOutput` → `ToolTextOutput` re-renders it through `JsonMessageContent` (a formatted key/value tree), so the DOM text no longer matches the raw `item.text` extracted here. A term that matches the raw source (quotes, `{`, escaped `\n`) is counted but `window.find` can't land on it in the rendered tree, and decoded text the tree shows may not be in the raw string. The same render-vs-segment gap applies to ANSI output (`ANSIDisplay` strips escapes) and to very large text truncated by `cappedText`.

This is pre-existing (the old whole-array `sanitizeStringify` had the same divergence), so nothing to fix in this PR — just confirming it's the intended limit of the "align match counts" goal rather than an oversight.

packages/inspect-components/src/transcript/eventText.ts
```diff
@@ -474,3 +476,59 @@ const extractContentText = (content: string | Array<Content>): string[] => {
...
+        ? [item.image]
+        : [];
+    case "document":
+      return "filename" in item && typeof item.filename === "string"
+        ? [item.filename]
+        : [];
+    case "audio":
+    case "video":
+      return [];
+    case "tool":
```
dragonstyle, line 527: Minor / dead-code question. The tool-event `result` schema (`generated.ts`) is `string | number | boolean | ContentText | ContentImage | ContentAudio | ContentVideo | ContentDocument | (ContentText | ContentImage | ContentAudio | ContentVideo | ContentDocument)[]`. So an array element can only be one of those five `Content*` types.

Given that, several branches here are unreachable for real event data:
- the `bigint` (line 497) and `symbol` (line 501) guards — JSON-sourced results can't produce these;
- this `case "tool"` recursion — a `{type:"tool", content:[...]}` wrapper is only synthesized at render time by `ToolCallView.normalizeContent`, never stored on the event.

Fine as defensive handling, but if it isn't deliberately guarding against the `as unknown as ToolEvent` casts / future schema drift, it's dead complexity worth trimming to match the schema.

Related: `data` content only reaches the `default` → `sanitizeStringify` → `<data />` branch, whereas the renderer (`ToolOutput.tsx:61`) shows `JSON.stringify(out.data)` as searchable text. That's a latent undercount if `data` ever appeared in a result array — it can't per schema today, and the old code collapsed it the same way, so not a regression. Worth a one-line comment noting `data` is intentionally excluded.

dragonstyle (approved):

PR #347, opened 2026-06-22
perf(viewer): lazy-load markdown-it-mathjax3 + batch markdown render commits

sjawhar:
Re-implements [a closed Inspect viewer perf change](https://github.com/UKGovernmentBEIS/inspect_ai/pull/3356), ported into `ts-mono` after the `_view/www` → submodule migration.

## What
Converts the static `markdown-it-mathjax3` import to a dynamic import so the heavy MathJax engine is fetched only when a transcript contains math, and keys the markdown render coordinator by a unique id so multiple `MarkdownDiv` instances rendering identical content no longer clobber each other's render/cancel bookkeeping.

## Verification
- Falsifiable chunk-split proof: the plugin-entry identifier is absent from the entry chunk and present in a separate dynamic chunk after the change.
- `pnpm --filter @tsmono/react test` + both app typechecks — green

One of a 7-PR series re-porting the closed Inspect viewer PRs into ts-mono.

### DIFF
--- a/packages/react/src/components/MarkdownDiv.test.tsx
+++ b/packages/react/src/components/MarkdownDiv.test.tsx
@@ -1,0 +1,33 @@
+// @vitest-environment jsdom
+import { render, screen, waitFor } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+
+import { MarkdownDiv } from "./MarkdownDiv";
+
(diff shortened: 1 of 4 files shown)


brandly:
i handled conflicts and reviewed in #422. thanks!

PR #354, opened 2026-06-22
transcript event navigation

Aprillion:
* addressing https://github.com/METR/hawk/issues/545 with MVP scope discussed with Charles
  * add keyboard shortcuts J/K (and down/up buttons for discoverability) + a link to an event page for inspecting long messages and/or API in a separate browser tab:
    image
  * H/L vim-inspired shortcuts for previous/next swimlane (subagents)
  * Ctrl+G for entering turn number manually


packages/inspect-components/src/transcript/TranscriptViewNodes.tsx
```diff
@@ -467,44 +358,208 @@ export const TranscriptViewNodes = forwardRef<
...
+    const settle = () => {
+      const el = container.querySelector<HTMLElement>(selector);
+      if (el) {
+        const delta =
+          el.getBoundingClientRect().top -
+          container.getBoundingClientRect().top -
+          offsetTop;
+        if (Math.abs(delta) > 2) container.scrollTop += delta;
+      }
+      if (frame++ < 8) raf = requestAnimationFrame(settle);
```
Aprillion: hm, I don't know how to test this - AFAICT this is _already broken on main_ of ts-mono when clicking a message link inside scout it just blinks a bunch and stays on top without scroll (and it does the same thing on this branch, but when I asked claude to delete the slop code it then restored it because it trusted agentic review more than me so I gave up and left the code as the agents want the code after arguing with them 3 times (they made some playwright tests that "prove" it works despite not being able to show me))
Image

FWIW the `?message=` works correctly in inspect log viewer on this branch with scroll-padding regardless of any `settle`:
  Image

* which seemed to have previously on main some random offset regardless of any requestAnimationFrame:
  Image

Aprillion:
fixed edge cases around swimlanes (subagents), de-slopified, added Ctrl+G shortcut to go to turn number and tested extensively => ready for external code review 🤞

image

dragonstyle:
Thanks for this - it is going to be great improvement to the user experience!

Here is feedback on the UI Side:

1) I think we should make the concept be a 'focus view' which the user can enter and exit (using the same affordance). Just clicking the focus mode button shouldn't open a new window, it should just navigate the current window. If the user ctrl + clicks we should open new window (mimicking browser click behavior). Note the slightly tweaked icon.

Screenshot 2026-07-02 at 3 06 09 PM

2) The focus mode needs to a gain a slightly stronger header to help contextualize the user.  Note the secondary bar which is fixed in the attached screenshot. It shows the transcript outline as a breadcrumb type nav on the left with the next/prev + exit focus mode control on the right. When we exit focus mode, we should be sure to return to the transcript at this event:

Screenshot 2026-07-02 at 3 05 42 PM

3) The current turn counter implementation is floating / absolute positioned which leads to some weird cases. I think we should try to just place it in the sticky header unless we really need it floating.

*clipped collapsed pills*
Screenshot 2026-07-02 at 2 57 59 PM

*content collision when scrolling*
Screenshot 2026-07-02 at 2 58 06 PM

*retries chip disappears in event view*
Screenshot 2026-07-02 at 1 44 55 PM
Screenshot 2026-07-02 at 1 44 52 PM

*other tabs show second event underneath their content*
Screenshot 2026-07-02 at 1 29 22 PM

dragonstyle:
Another note - this PR appears to have lost the background color and border on sticky headers. I think we'd like to keep that.

Screenshot 2026-07-02 at 3 24 16 PM

dragonstyle:
For the jump to turn portion of the nav, I put together a handoff for Claude Code:

[Transcript jump to turn feature.zip](https://github.com/user-attachments/files/29613681/Transcript.jump.to.turn.feature.zip)

which makes the jump UI a centered popup box like:

Screenshot 2026-07-02 at 5 05 57 PM

Screenshot 2026-07-02 at 5 06 03 PM

apps/inspect/src/app/samples/transcript/TranscriptPanel.tsx
```diff
@@ -287,6 +310,49 @@ export const TranscriptPanel: FC<TranscriptPanelProps> = memo((props) => {
     [builder, urlLogPath, urlSampleId, urlEpoch, logFile, logDir]
   );
 
+  // Hash-route href for the open-in-new-tab control. Relative `#…` so a normal
+  // ctrl/cmd-click or middle-click opens it in a new tab of the same SPA. The
+  // VS Code webview has no browser-tab model (and can't open one), so the
+  // control is hidden there by returning no URL.
+  const getEventFocusUrl = useCallback(
```
dragonstyle: I think this should become something more public / shared out in url.ts or related

packages/inspect-components/src/chat/ChatViewVirtualList.tsx
```diff
@@ -57,12 +57,18 @@ const ChatItem = ({ children, ...props }: VirtualListItemProps) => {
 
 const chatComponents = { Item: ChatItem };
 
+// Scroll-padding applied to the message list's scroll-to-index (deep-link,
+// find, initial message): a small negative value lands the target snug at the
+// top instead of tucked a row too high. Empirically tuned for the chat surface;
+// the transcript has its own (kTranscriptScrollPaddingStart). NOT derived from
+// the host's chrome height - the chat owns this as a constant.
+const kChatScrollPaddingStart = -15;
```
dragonstyle, line 61: this seems a little sketchy - if you have verified this works in all areas of usage this works, it might be ok?

packages/inspect-components/src/transcript/event/EventPanel.tsx
```diff
@@ -110,15 +108,46 @@ export const EventPanel: FC<EventPanelProps> = ({
...
-    }
+  const [storedNav, setStoredNav] = useProperty(eventNodeId, "selectedNav", {
+    defaultValue: defaultPillId,
+  });
+
+  // On the single-event focus page a shared tab (by NAME) is remembered across
+  // events, so navigating between turns keeps the same tab open instead of each
+  // event reverting to its own per-event selection. Absent in the main
+  // transcript, where each panel keeps its own `selectedNav`.
+  const focusTab = useContext(FocusTabContext);
```
dragonstyle: Seems a little odd that something as general as the eventpanel would 'know' anything about focus tab presentation. It seems better to extract this to the host and pass as an option or something to make this more decoupled.

packages/inspect-components/src/transcript/event/EventPanel.tsx
```diff
@@ -153,114 +179,126 @@ export const EventPanel: FC<EventPanelProps> = ({
...
+          <span className={clsx("text-style-secondary", styles.label)}>
+            {text}
+          </span>
+        ) : null}
+      </div>
+      {eventLabel && (
+        <MessageLabel label={eventLabel} className={styles.eventLabel} />
+      )}
+    </div>
+  );
```
dragonstyle: A lot of churn in this control - is it all really necessary?

packages/inspect-components/src/transcript/timeline/components/TimelineSwimLanes.tsx
```diff
@@ -821,6 +830,30 @@ const HeaderRow: FC<HeaderRowProps> = ({
           {rootDisplay}
         </button>
       )}
+      {laneNav && (
```
dragonstyle: I had trouble getting this lane nav to show. Where should I be seeing it?

packages/inspect-components/src/transcript/ToolEventView.tsx
```diff
@@ -75,10 +75,6 @@ export const ToolEventView: FC<ToolEventViewProps> = ({
   const displayName = resolvedView?.title || title || name;
   const panelTitle = displayName ? `Tool: ${displayName}` : "Tool";
 
-  const turnLabel = context?.turnInfo
```
dragonstyle, line 78: if we move away from the global turn label, this might need to come back.

packages/inspect-components/src/transcript/TranscriptViewNodes.tsx
```diff
@@ -4,36 +4,37 @@
  * scroll-to-event/index. Apps provide collapse state via callback props.
  */
```
dragonstyle, line 6: I realize this may be related to your other comments, but is all this churn also necessary?

packages/react/src/hooks/useListKeyboardNavigation.ts
```diff
@@ -10,16 +10,94 @@ interface ListKeyboardNavigationOptions {
   scrollRef?: RefObject<HTMLDivElement | null>;
   /** Total number of items in the list. */
   itemCount: number;
+  /** `j` — scroll to the next turn (down, vim-style). */
+  onNext?: () => void;
+  /** `k` — scroll to the previous turn (up, vim-style). */
+  onPrev?: () => void;
+  /** `h` — move to the previous agent lane (left, vim-style). */
+  onPrevAgent?: () => void;
```
dragonstyle: These seem very specific to appear in 'list navigation'.

dragonstyle (changes requested):
Some code feedback as well. Thanks for PR!

Aprillion:
sticky headers consolidated, Ctrl+G for go to turn, f for focus view in the same window (including in vs code extension, screenshot from retesting various color themes):
image

...I asked Fable to fix multiple issues as I was retesting, some of them turned out to be not caused by this branch so I asked it to split the most adjacent stuff into separate commit in case you'd like to move it out of this PR 😰

packages/react/src/hooks/useListKeyboardNavigation.ts
Aprillion: moved

packages/inspect-components/src/transcript/TranscriptViewNodes.tsx
Aprillion: whenever I tried to delete/revert anything and/or ask Fable, more edge cases popped up, so in the end I asked it to use its own judgement to keep what's necessary and remove what's not and then I retested everything - sorry if it's still too much churn (probably even more after more testing), but AFAICT it seems necessary to make everything scroll-navigation related (turns + return from focus mode + left sidebar + find + search + swimlanes) work together

packages/inspect-components/src/transcript/ToolEventView.tsx
Aprillion: yup

packages/inspect-components/src/transcript/timeline/components/TimelineSwimLanes.tsx
Aprillion: now displayed as disabled even if there are no subagent swimlanes ("for discoverability" according to Fable and I agree) .. but meaningful only for transcripts with subagents:

image

packages/inspect-components/src/transcript/event/EventPanel.tsx
Aprillion: I agree, but after 1 more agentic cleanup, most things I tried to poke turned out necessary as far as I could tell before I got tired from retesting regressions (see my other comment)

packages/inspect-components/src/transcript/event/EventPanel.tsx
Aprillion: FocusTabContext deleted; generic getSelectedTab/onSelectTab on EventPanelCallbacks; focus page supplies them from URL ?tab=.

packages/inspect-components/src/chat/ChatViewVirtualList.tsx
Aprillion: yes it is sketchy 🙈  but verified everywhere I could - I poked Claude until it gave me repro steps for all usages it could trace, e.g.:
image

apps/inspect/src/app/samples/transcript/TranscriptPanel.tsx
Aprillion: Moved to useSampleEventFocusUrlBuilder() in url.ts

dragonstyle:
> I asked Fable to fix multiple issues as I was retesting, some of them turned out to be not caused by this branch so I asked it to split the most adjacent stuff into separate commit in case you'd like to move it out of this PR

Thanks for finding and fixing! I think a separate PR would be good. Eric is currently working on revamping collection data loading right now, so I anticipate a lot of churn in the code in the next couple of days. It would be good to sit on the new PR for a few days to see where Eric lands.

dragonstyle:
This is shaping up to be a great feature. Some additional things I ran into testing:

**Loading Sample Data**
- Formatting of the 'loading same data' is inconsistent with other loading data formatting.
- Ideally would show the top navbar with an activity bar animating underneath it and loading sample data placed below in small font with animated elipsis

**Navbar**
- Back button in top nav of focused view seems like it should exit focus view (that was the previous state). If you feel strongly it should return to the parent of the sample view, check to be sure it goes to the view that originated the sample view (e.g. it could be launched from a logview or directly from the samples list, we should return to the proper parent).

- Similarly, the home button needs to return to the original parent (log file or samples list).

**Goto Turn**
- The goto dialog needs additional padding (and textbox should also probably be smaller). 
Screenshot 2026-07-06 at 10 11 38 AM

- for long turn counts, the prefill value in the textbox can be truncated
Screenshot 2026-07-06 at 10 12 54 AM

**Focus View**
- When navigating into a subagent, the left side of focus view only shows 'main' rather than the path to the subagent (e.g. main > subagent).

- The turn counts seem sometimes inconsistent. In petri transcripts I am on turn 1/12 and then i go into focus view and it shows turn 14/25.

- Can you push the padding / margin out to right/left align elements (see screenshot)

Screenshot 2026-07-06 at 10 08 53 AM

dragonstyle:
Add a Fable list of findings:

### Confirmed bugs

1. [VirtualList.tsx:586](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/packages/react/src/virtual/VirtualList.tsx#L586) — extended-find permanently latches the auto-scroll guard, killing scroll persistence for the rest of the mount. searchInData cancels an in-flight settleScrollToIndex via cancelAnimationFrame(settleFrameRef.current) but never calls finish(), violating the settle loop's own documented invariant ("every exit path must release it"). A settle runs up to ~30 frames after any deep link or outline jump; a find hit landing in that window leaves isAutoScrollingRef true forever, so persistOnScroll returns early on every subsequent user scroll — position is never recorded again and is lost on tab flips.

2. [FocusTurnView.tsx:255](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/packages/inspect-components/src/transcript/FocusTurnView.tsx#L255) — focus mode never resets scroll when navigating turns. Turn/lane navigation is just setParam("event", …) re-rendering the slice into the same overflow-y: auto container; nothing writes scrollTop, there's no key remount, and initialEventId isn't passed. Scroll partway down a long turn and press j — you land mid-way (or at the bottom) of the next turn instead of at its top. This is the core interaction of the new page.

3. [useTranscriptKeyboardNavigation.ts:32](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/packages/inspect-components/src/transcript/hooks/useTranscriptKeyboardNavigation.ts#L32) — window-capture j/k/gg/G handlers (and GoToTurnBar's Ctrl+G) leak into transcript preview surfaces. The hook attaches a capture-phase window listener with no container-visibility check, and TranscriptViewNodes always defines onNext/onPrev. Pre-existing preview hosts ([scanReferencePreviews.tsx:21](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/apps/inspect/src/app/samples/scans/scanReferencePreviews.tsx#L21), [refs.tsx:245](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/apps/scout/src/app/utils/refs.tsx#L245)) silently inherit global handlers: with a main transcript and a preview both mounted, stopImmediatePropagation makes it first-registered-wins, and the main panel's listener re-registers on dep churn (streaming, collapse) and falls behind — j/k then scrolls the tiny hover preview instead. GoToTurnBar's checkVisibility guard passes for a visible popover, so Ctrl+G can open "Go to turn 1/1" inside a hover preview.

4. [TranscriptViewNodes.tsx:361](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/packages/inspect-components/src/transcript/TranscriptViewNodes.tsx#L361) — ?message= deep links land the message behind the pinned event header. The new settle scrolls the message top to exactly offsetTop — the same line where EventPanel's opaque, z-index-1 sticky header pins. The old code measured the pinned header (measureStickyBottom()) and added 20px; that compensation was dropped, so the message's first ~30px render underneath the header. The settle budget also shrank from ≤90 frames with a stability check to a fixed 9 frames, so slow markdown/image reflow can leave the target mis-positioned.

5. [useScrollTrack.ts:67](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/packages/react/src/hooks/useScrollTrack.ts#L67) — events in the final viewport can never become "current" except the last one. The old sliding detection point (docstring: "so every item can be selected") was replaced by a binary top-crossing check plus a last-visible fallback gated on being within 24px of max scroll. Any element starting in the last viewport-height of content is now unreachable as "current": scrolling to the bottom, the outline highlight and j/k turn tracking jump from an earlier event straight to the last one, permanently skipping those in between. Comments suggest a deliberate tradeoff for jump-landing consistency, but the intermediate-element gap looks unintended.

6. [TranscriptViewNodes.tsx:564](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/packages/inspect-components/src/transcript/TranscriptViewNodes.tsx#L564) — first j skips turn 1 when pre-turn content is on screen. currentTurnIndex initializes to 0 ("turn 1 current") and onTopEvent defaults unmapped pre-turn rows to turn 1 (?? 1), so with init spans/info events occupying the initial viewport, the first j computes goToTurn(0 + 1) and lands on turn 2 — turn 1's anchor is never visited. Needs a "before first turn" sentinel (e.g. −1).

7. [GoToTurnBar.tsx:94](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/packages/inspect-components/src/transcript/GoToTurnBar.tsx#L94) — Ctrl+G's inline editable-target check misses shadow-DOM inputs and `select` tag. The handler hand-rolls instanceof HTMLInputElement/HTMLTextAreaElement/isContentEditable instead of calling isEditableTarget from @tsmono/util — the helper this same PR extended with shadow-root piercing precisely for VS Code webview controls (<vscode-textarea> retargets keydowns to the shadow host, which fails all three checks). Typing Ctrl+G in such a field hijacks the chord and opens the bar. useTranscriptKeyboardNavigation and FocusTurnView already use the helper.

8. [useFocusTurnNavigation.ts:190](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/packages/inspect-components/src/transcript/hooks/useFocusTurnNavigation.ts#L190) — breadcrumb can navigate to the wrong agent instance. computeAgentLanes filters out lanes with no direct model turns, and laneCrumbs matches ancestors by name-path equality (last prior match). When the true enclosing agent has no direct turns and an earlier sibling with the same name does (common orchestrator/dispatch pattern), the crumb becomes clickable and goToLane navigates to the earlier instance's first turn instead of degrading to a plain label as documented.

### Verified secondary issues (performance / cleanup)

- Hot-path waste: [useFocusTurnNavigation.ts:115](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/packages/inspect-components/src/transcript/hooks/useFocusTurnNavigation.ts#L115) re-runs a full O(n) flatTree flatten on every keypress despite having the memoized flat two lines up; [useScrollTrack.ts:51](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/packages/react/src/hooks/useScrollTrack.ts#L51) rebuilds a Set of every flattened node id (thousands) ~10×/sec while scrolling — hoist to a memo; [TranscriptViewNodes.tsx:330](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/packages/inspect-components/src/transcript/TranscriptViewNodes.tsx#L330) currentTurnIndex state is never rendered (only mirrored into a ref), so every turn-boundary crossing during scroll re-renders the component for nothing — a plain ref suffices and also removes the stale-read workaround in goToTurn; [useStickyStuck.ts:90](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/packages/inspect-components/src/transcript/hooks/useStickyStuck.ts#L90) every panel header now registers (previously only turn-labeled model headers) and each registration synchronously rect-reads all registered headers — O(k²) layout reads per virtual-list mount batch; use the rAF-coalesced scheduleUpdate instead.

- Duplication: modelEventTabNames() ([ModelEventView.tsx:325](vscode-webview://17t07ltilu9p87jqhk8hiv7iei6f3bvea592rk4nketa52lesamf/src/inspect_ai/_view/ts-mono/packages/inspect-components/src/transcript/ModelEventView.tsx#L325)) is exported but has zero callers and hard-codes a tab list that must mirror the JSX — delete it. onOpenEventFocus is plumbed through both EventNodeContext and EventPanelCallbacks while sibling getEventFocusUrl is a plain prop — pick one channel. The agent-lane depth-stack walk now exists three times (computeAgentLanes, resolveEventTurnAnchor, computeLaneFirstAnchors), and the ?event=/?message= URL-writing callbacks are copy-pasted verbatim between the two apps (4 sites). useLoadSampleFromRoute says it "mirrors LogSampleDetailView's init sequence" but that component still keeps its own inline copy — convert it too.

- Plausible but unconfirmed: goToPrevFrom's "did the scroll move" probe (2 rAFs, 24px tolerance) races the settle loop (up to 30 frames of re-jumps on heavy content) — k can be swallowed or skip a turn; keying the decision on settle completion (it already has an onDone) would be robust. Also, focus mode mounts no StickyScrollProvider, so useStickyStuck is a silent no-op there (minor cosmetic: pinned headers never get the stuck band).

dragonstyle (changes requested):
Another round of code and UX feedback...

Aprillion:
Agentic fixes were not up to my quality standards and I was a bit distracted, but today I finished manual testing and helped Fable decide to reduce scope what to keep in this PR for turn navigation - all human findings fixed (and re-tested), agentic findings that were reproducible should be also fixed (but retested with less care and while the latest repeated review from 5.6 Sol looked plausible, I decided to stop the whackamole for now unless requested by Charles as priority for this PR and not as follow-up investigation)...

full report in https://peter.hozak.info/claude/pr354-transcript-navigation.html

dragonstyle (changes requested):
this is looking so good! A couple more nits:

- I'm seeing the outline number of turns when navigating using j/k not being bold (it flashes bold and then unbolds. We should try to have the turn entry bolded whenever we're within its scope of turns (I can better explain if you need me to - in the below, I'd expect '9 turns' to be bold)

Image

- I think the next/previous agent buttons are a little random in the swim lanes breadcrumbs area (it's not clear what they control). I think it would be better without them here (and I'd vote for out of the focus view too).

Image

- Could we make the switcher a pill as in the full transcript?

Image

dragonstyle (changes requested):
One more pass with Fable and a few real issues I think (they're getting small, so I'm guessing this will be the last round). I took it for a spin before your most recent changes it is looking great.

dragonstyle (approved):

PR #355, opened 2026-06-23
Enable lint rules for Inspect

brandly:
We started using shared rules in #320, but some rules were turned `off`. This PR enables more rules and handles the fallout.


apps/inspect/src/setupTests.ts
```diff
@@ -14,6 +14,8 @@ global.TextEncoder = TextEncoder;
 // Polyfill structuredClone for Node.js versions that don't have it
 if (typeof global.structuredClone === "undefined") {
   global.structuredClone = (obj) => {
+    // Polyfill clones via a JSON round-trip, which is inherently untyped.
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
brandly: I don't think we need this polyfill at all. [It's supported in node 17](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone#browser_compatibility) and we say `>= 22`: https://github.com/meridianlabs-ai/ts-mono/blob/7aeda758d7f2b768272e02a69b97a05101850b22/package.json#L7

PR #357, opened 2026-06-23
perf(inspect): cold-open latency — promise-memo open, tail cache, un-gated sample fetch

kaifronsdal:
## Summary

Three changes that reduce cold-open latency for sample URLs in the inspect viewer. Profiled against 24 diverse `.eval` files (4 KB – 513 MB; 1 – 6,960 samples; deflate + zstd) plus real-browser traces (Firefox + Chrome).

| change | what the data showed | effect |
|---|---|---|
| **Promise-memoise `remoteEvalFile`** | Browser traces: `log-info`/EOCD/cdir fetched ×2–×4 on every cold nav (137 duplicate requests across 18 cases) because concurrent callers each ran their own `openRemoteLogFile`. | Concurrent reads share one open; removes a full RTT chain. |
| **128 KB tail-window cache in `openRemoteZipFile`** | EOCD + ZIP64 + cdir + `header.json` all live in the file tail (verified across the corpus; cdir fits for 22/24 files). | `info → EOCD → cdir` collapses to one network read. Existing parsing code unchanged — the cache is a transparent `fetchBytes` shim. |
| **Prefetch the sample alongside summaries** | `loadSample` is gated on `logSelection.sample`, which only resolves after `summaries.json` (or the journal fallback) loads — even though `readSample` only needs `id`+`epoch`, both of which are known from the route. | A separate effect calls `warm_log_sample(file, id, epoch)` as soon as the route handle resolves; the unchanged `loadSample` flow then resolves from the warmed promise instead of issuing a second fetch. The sample download runs `‖` summaries instead of after. `retryUncached=false` on the warm avoids a wasted zip reopen when the sample isn't in the cdir yet (running eval). |

(An earlier revision rewrote `useLoadSample` to fire speculatively from the route handle and tracked the miss/retry state with a `useRef`. Per review feedback that's been replaced with the cache-warm approach above — `useLoadSample.ts` is now +27/−0, one isolated effect, and the load logic itself is untouched.)

Measured at ~150 ms RTT: cold-open depth 5 → 3 (~1022 → ~600 ms). Helps all deploy modes (view-server, static-http, VS Code).

## Tests

- `remoteZipFile.test.ts` (new): builds a minimal zip fixture, drives `openRemoteZipFile` with a recording `fetchBytes`, asserts (a) zip-open + in-window `readFile` = 1 network call, (b) out-of-window entry = 2 calls, (c) Range-ignoring server detected.
- `client-api.test.ts`: concurrent `get_log_details` calls share one `openRemoteLogFile`; `cached=false` reuses an in-flight open but re-opens once resolved; a rejected open is not cached. New `warm_log_sample` block: warmed hit returns without a second read; warmed miss falls through; `edit_log` clears the warm.
- All 487 existing tests pass; lint + typecheck clean.

## Notes

- The 128 KB window is a prefetch hint, not a correctness assumption. If the cdir is larger (>~1 k entries), the existing exact-range fetch fires unchanged — depth degrades gracefully to 3 instead of 2.
- Earlier revisions of this PR also added a `parallelChunkSize` option for the presigned-S3 path; that's been dropped along with the companion `--direct-urls` CLI flag (UKGovernmentBEIS/inspect_ai#4337) — the throughput problem it targeted is better solved by Dev Tunnels.

### DIFF
--- a/apps/inspect/src/client/api/client-api.test.ts
+++ b/apps/inspect/src/client/api/client-api.test.ts
@@ -1,6 +1,9 @@
 import { describe, expect, test, vi } from "vitest";
 
-import { openRemoteLogFile } from "../remote/remoteLogFile";
+import {
+  openRemoteLogFile,
+  SampleNotFoundError,
+} from "../remote/remoteLogFile";
 
 import { clientApi } from "./client-api";
 import {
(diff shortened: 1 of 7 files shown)


dragonstyle (commented):
I'm most concerned about the complexity of ungating the useLoadSample. Rather than speculatively fetch, could we treat the sample read as a speculative cache warm, but keep the old load flow but against a warm cache? I'm thinking this might simplify the logic (the useRef + the churn around all the different states is concerning).

kaifronsdal:
Reworked per the review comment — `useLoadSample` is back to main + one isolated warm effect; the speculative-miss ref/state is gone. The prefetch now lives in `client-api` as a single-slot promise cache (`warm_log_sample`). Net diff to `useLoadSample.ts` is +27/−0. Rebased on main; 487 tests green.

kaifronsdal:
One known tradeoff with the cache-warm approach: the warm call has no `onProgress`, so the determinate download bar doesn't show on a cold open. In practice `onProgress` only ever fired for samples > 8 MB compressed (the `PARALLEL_CHUNK_THRESHOLD` in `fetchBytesParallel`), and then only at 8 MB granularity — smaller samples already showed the indeterminate sweep on main. Wall-clock is shorter (the download starts earlier and runs in parallel with summaries), so the net is a shorter indeterminate sweep vs a slightly longer chunky determinate bar for the rare huge-sample case. Can wire late-bound progress if it matters, but didn't seem worth the extra state.

dragonstyle:
1. `apps/inspect/src/client/api/client-api.ts:105`
cached=false now reuses any in-flight cached open, but openRemoteLogFile() has already issued api.get_log_info() as soon as that promise exists (remoteLogFile.ts:95). That means a fresh syncLog / refreshLog call can reuse a size/cdir snapshot from before the fresh call, so running evals can miss just-flushed samples/details. I’d avoid sharing in-flight cached opens for cached=false, or make the freshness tradeoff explicit and bounded.

2. `apps/inspect/src/client/api/client-api.ts:223`
A real get_log_sample(..., onProgress) consumes the warmed promise before it can attach its progress callback. Since warm_log_sample() starts the fetch with onProgress undefined (client-api.ts:284), large cold-open sample downloads now look like a plain loading state until completion, even though useLoadSample.ts:118 still sets up progress. This regresses the slow path the PR targets.

dragonstyle:
Closing this one. Since June the data layer moved to react-query (#389) and `useLoadSample.ts` no longer exists, so the promise-memo and prefetch parts would need rebuilding on the new layer rather than rebasing, and the open question from 6/27 about the ungated sample fetch is moot in that design.

The tail-window cache in `remoteZipFile.ts` is a separate, self-contained win that still applies cleanly to `main`. Please open a PR with just that part (the `remoteZipFile.ts` change and its test) and I'll take it right away.

PR #358, opened 2026-06-24
Enable lint rules in Scout

brandly:
Similar to #355, we'll also enable these rules for Scout and handle the fallout.


apps/scout/src/app/server/useServerTranscripts.test.ts
```diff
@@ -100,7 +100,7 @@ describe("useServerTranscripts", () => {
     });
 
     expect(capturedBody).toMatchObject({
-      filter: JSON.parse(JSON.stringify(filter)),
```
brandly, line 103: use `deepCopy` and avoid the cast?

PR #362, opened 2026-06-25
Prevent automatic loading of untrusted media

ericwinsor-aisi:
## This PR contains:
- [ ] New features
- [ ] Changes to dev-tools e.g. CI config / github tooling
- [ ] Docs
- [x] Bug fixes
- [x] Code refactor

### What is the current behavior?

Markdown image syntax is rendered as an ``. Typed audio and
video create `<source src="...">` elements, and image documents create an
``.

When the viewer displays this content, the browser automatically requests those
URLs. No click or other user interaction is required.

This allows model-, tool-, or scanner-controlled content to make the reviewer's
browser contact a chosen external, same-origin, local, or private-network
address.

### What is the new behavior?

Markdown images no longer create media elements. Absolute HTTP and HTTPS
destinations are shown as visible external links and load only after a click.

Typed media previews only when it contains validated base64 data with a MIME
type matching the media type. HTTP and HTTPS references become visible external
links, while other references remain text.

### What rendering policy does this establish?

> Rendering untrusted content does not perform network or filesystem access.
> Automatic media rendering is limited to media bytes already present in an
> approved inline representation. Remote media requires a separate, explicit
> user action.

A typed media block describes how content should be displayed. It does not give
the content permission to make a browser request.

The implementation establishes this policy in shared code:

- one parser handles data URIs and absolute HTTP/HTTPS URLs;
- one media policy validates image, audio, video, and document sources;
- Markdown image rendering never creates an image element;
- shared media components use the same validation rather than separate prefix
  checks; and
- links opened in a new tab use `noopener noreferrer`.

### Does this PR introduce a breaking change?

Remote media and Markdown images no longer preview automatically. Their
destinations remain available as links or text. Valid inline image, audio, and
video data continues to render normally.

### Other information:

The policy is shared by Inspect and Scout.

Testing includes unit tests, typecheck, lint and formatting checks, production
builds, and browser tests confirming that media URLs receive no request before
an explicit click.

### DIFF
--- a/packages/react/src/components/renderedHtmlSanitizer.ts
+++ b/packages/react/src/components/renderedHtmlSanitizer.ts
@@ -16,11 +16,14 @@ const FORBIDDEN_TAGS = [
   "foreignobject",
   "form",
   "iframe",
+  "image",
+  "img",
   "input",
   "link",
   "meta",
   "mpath",
   "object",
+  "picture",
   "script",
   "select",
   "set",
@@ -101,8 +104,7 @@ const UNSAFE_CSS_PATTERN =
   /@import|behavior\s*:|binding\s*:|expression\s*\(|javascript\s*:|vbscript\s*:|url\s*\(/i;
 
 const PURIFY_CONFIG: Config = {
-  ADD_ATTR: MATHJAX_ATTRS,
-  ADD_DATA_URI_TAGS: ["img", "image"],
+  ADD_ATTR: [...MATHJAX_ATTRS, "target"],
   ADD_TAGS: MATHJAX_TAGS,
   ALLOW_DATA_ATTR: true,
   ALLOW_UNKNOWN_PROTOCOLS: false,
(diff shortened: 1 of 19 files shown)


packages/react/src/components/renderedHtmlSanitizer.ts
```diff
@@ -179,12 +181,22 @@ const installHooks = (purify: DOMPurifyInstance): void => {
...
       hookEvent.keepAttr = false;
       node.removeAttribute(hookEvent.attrName);
     }
   });
+
+  purify.addHook("afterSanitizeAttributes", (node) => {
+    if (
+      node instanceof Element &&
+      node.tagName.toLowerCase() === "a" &&
+      node.getAttribute("target") === "_blank"
```
epatey, line 195: This should probably `.toLowerCase()` as well.

epatey:
@ericwinsor-aisi, thanks for this. I may follow up, if/when I have time, to enhance the UX to have a "Some media/images are not displayed..." banner like gmail does to allow the user to opt into the nicer rendering.

PR #366, opened 2026-06-25
Implement transcript search methods in the vscode API client

brandly:
The Search rail button in the inspect log viewer is gated on the API client exposing list_searches/post_search/get_search_result (via useInspectSearchContext). The vscode client omitted all three, so the button never appeared when embedded in VS Code -- neither the panel nor the "install Scout" fallback was reachable.

Add the three methods, proxied over the existing JSON-RPC client to the extension (which forwards to inspect_ai's /scout/* endpoints). Follow the established patterns: accept both wire shapes, map kJsonRpcMethodNotFound to a clear "newer extension required" error, and treat 404/empty results from get_search_result as null (view-server polling parity).

These changes in addition to [some changes in the extension](https://github.com/meridianlabs-ai/inspect_vscode/pull/121) will put Inspect in a position to use Scout's search functionality when embedded in vscode.

### DIFF
--- a/apps/inspect/src/client/api/vscode/api-vscode.ts
+++ b/apps/inspect/src/client/api/vscode/api-vscode.ts
@@ -6,6 +6,10 @@ import {
   LogFilesResponse,
   LogInfo,
   LogUpdate,
+  Result,
+  SearchInputListResponse,
+  SearchRequest,
+  SearchResponse,
 } from "@tsmono/inspect-common/types";
 import {
   getVscodeApi,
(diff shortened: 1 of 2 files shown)


brandly:
Easy changes to pull in I think, but this is throwaway code once we start [using the more generic transport](https://github.com/meridianlabs-ai/ts-mono/pull/371)

dragonstyle (approved):

dragonstyle:
@brandly sorry merged the newer one first - merge conflict to resolve...

PR #371, opened 2026-06-26
Generic HTTP transport for Inspect in vscode

brandly:
https://github.com/meridianlabs-ai/inspect_vscode/pull/122 opens the door for Inspect's viewer to send generic `http_request` messages over the JSON-RPC proxy. This PR starts taking advantage of those changes.

Those vscode changes lift functionality that was already present for Scout and expose it to Inspect. Similarly, this PR lifts logic out of Scout and integrates it into Inspect. Since the vscode extension and the Scout viewer are not kept in sync, now the Inspect viewer will check to see if the `http_request` transport is available and take advantage of it if so. The old transport logic is left in place and will be used if `http_request` is not available. 

By default, vscode extensions will auto-update [two hours after publish](https://thehackernews.com/2026/06/vs-code-adds-2-hour-extension-auto.html), so I don't think we'll have to keep the other transport in place for long.


dragonstyle:
Fixes #374

PR #373, opened 2026-06-27
Constrain automatic log loading to trusted locations

ericwinsor-aisi:
### Current behavior

The viewer used the same strings for trusted deployment configuration and URL- or message-selected log locations. A crafted query, hash route, or window message could therefore replace the static log target and cause automatic GET, HEAD, or range requests to same-origin endpoints, local/private-network services, or external origins.

Supporting hosted absolute log URLs is intentional. Treating receipt of a viewer link as authority to contact any location is not.

### New behavior

Resolve transport and initial log authority only from trusted embedded configuration or the VS Code host. Query parameters, routes, listings, restored state, and host updates now select through a shared exact-file or canonical-directory policy.

Every browser-direct log request is guarded by a live runtime capability, including JSON GETs, manifests, auxiliary files, and `.eval` HEAD/range requests. Requests use same-origin credentials only, omit the referrer, and reject redirects that could change the approved destination.

A broader HTTP(S) file or directory can still be opened after a visible action approving that exact location for the current page. Unsupported, credential-bearing, protocol-relative, `file:`, `data:`, `blob:`, and custom-scheme locations remain non-loadable.

### Behavioral decisions

- Keep automatic loading for exact files and canonical directories configured by the publisher or VS Code host.
- Keep bundled and hosted relative selection automatic only within the declared scope.
- Do not let `log_file`, `log_dir`, `task_file`, hash routes, restored state, or `inspect_server=true` choose or widen the transport.
- Keep one in-memory ad hoc grant and revoke it when the candidate changes; do not add persistent or host-wide trust.
- Treat manifest entries and server listings as data: reject escaping, absolute, ambiguous, or duplicate-canonical browser entries, and do not let cached or browser-derived handles establish server authority.
- Keep view-server and VS Code paths on their separately scoped transports. Normal browser pages ignore VS Code state and runtime log-update messages.
- Hide server-only edit, search, and download affordances for browser-hosted ad hoc logs.

### Compatibility

Generated bundles, embedded directory roots, nested relative hash links, exact publisher-configured hosted logs, standalone server routes, VS Code selections, JSON logs, and range-readable `.eval` logs retain their declared behavior.

Query-only arbitrary remote links now require the explicit action. Publishers that need automatic hosted loading can embed the existing trusted log configuration.

### Testing

- Inspect typecheck and lint
- Inspect unit tests: 529 passed
- Inspect production build
- Util typecheck, lint, and tests: 91 passed
- Playwright end-to-end suite: 83 passed
- `git diff --check`

### DIFF
--- a/apps/inspect/e2e/fixtures/handlers.ts
+++ b/apps/inspect/e2e/fixtures/handlers.ts
@@ -28,6 +28,11 @@ export const defaultHandlers = [
     return HttpResponse.json({ log_dir: "/home/test/logs" });
   }),
 
+  // Legacy combined log root
+  http.get("*/api/logs", () => {
+    return HttpResponse.json({ logs: [], log_dir: "/home/test/logs" });
+  }),
+
   // Log file listing
   http.get("*/api/log-files*", () => {
     return HttpResponse.json({
@@ -48,6 +53,11 @@ export const defaultHandlers = [
 
   // Flow (optional, 404 is acceptable)
   http.get("*/api/flow*", () => {
+    return new HttpResponse(null, { status: 404 });
+  }),
+
+  // No live samples in the default completed-log fixture
+  http.get("*/api/pending-samples*", () => {
     return new HttpResponse(null, { status: 404 });
   }),
 ];
--- a/apps/inspect/e2e/security-log-selection.spec.ts
+++ b/apps/inspect/e2e/security-log-selection.spec.ts
@@ -1,0 +1,386 @@
+import { http, HttpResponse } from "msw";
+
+import { expect, test } from "./fixtures/app";
+import { createEvalLog } from "./fixtures/test-data";
+
+declare global {
+  interface Window {
+    __LOG_SELECTION_FETCHES__?: string[];
+  }
+}
+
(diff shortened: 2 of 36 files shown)


dragonstyle:
Thanks for this. The separation you drew between "who may set the log location" and "who may only propose one" is the model we'd keep, and the fetch hardening (no referrer, same-origin credentials, no redirects) is something we should have regardless.

I'm going to close the PR itself. It has been open since late June and the viewer's bootstrap has been rebuilt underneath it: backend and initial-log resolution now live in `apps/inspect/src/app_config/` (#482) and the data layer moved to react-query (#389), so 23 of the 36 files conflict and the wiring would need to be redone at the new seam rather than rebased. I don't want to ask you to redo that speculatively.

So the concern isn't lost, I've opened https://github.com/meridianlabs-ai/ts-mono/issues/615 describing the gap on current `main` and pointing back here for the policy design and tests. If you'd like to take a fresh, smaller run at it against `app_config/resolveAppConfig()`, I'd welcome it; otherwise we'll pick it up from the issue.

PR #378, opened 2026-06-30
Prevent white flash on dark-theme load

Aprillion:
no more FOUC on manual reload of viewers from `pnpm dev` 👀🔦

fixed while working on https://github.com/meridianlabs-ai/ts-mono/pull/354 because it drove me crazy (HMR was not reliable so I reloaded _a lot_), but it was out of scope of that feature, so separate PR


brandly (approved):

brandly:
thanks @Aprillion!

PR #384, opened 2026-07-01
Consolidate json-worker, eliminate double parse, add real-payload benchmarks

dragonstyle:
## Summary

Two near-identical copies of `json-worker.ts` existed (`packages/util` and `apps/inspect`, each embedding its own 42KB base64 JSON5 bundle). This PR consolidates to a single copy in `@tsmono/util`, fixes the large-payload performance problems, and adds a benchmark harness with committed before/after results — including runs against ten real eval-log, scout-transcript, summary, and SWE-bench-style files from 7MB to 186MB.

## Consolidation

- `packages/util/src/json-worker.ts` is the single copy, adopting the inspect superset (`asyncJsonParseBytes`, typed worker responses, sync `jsonParse` with JSON5 fallback)
- Deleted `apps/inspect/src/utils/json-worker.ts`; its six import sites now use `@tsmono/util`
- JSON5 fallback is unchanged and universal: every path tries `JSON.parse` first and falls back on failure

## Performance

**Double parse eliminated.** The worker previously `JSON.parse`d large payloads, `JSON.stringify`d the result, transferred the bytes, and `JSON.parse`d them *again* on the main thread — parsing everything twice with the second parse on the thread the worker exists to protect. The main thread also paid a full `TextEncoder.encode` pass before every dispatch; strings now cross to the worker as a plain postMessage clone.

**Shape-adaptive result transport.** Structured clone cost scales with node count, not bytes. On a real 186MB scout transcript-events file (~36M nodes, ~5 bytes/node), cloning the parsed graph blocked the main thread longer than a plain `JSON.parse` of the source. The worker probes structural density (separators per char over a 16MB state-correct window, skipping string contents via indexOf hopping — ~10–45ms off-thread) and, for node-dense strict-JSON payloads over 10MB, signals the main thread to run a single `JSON.parse` instead of cloning — the cheapest possible materialization. String-heavy payloads (eval logs, tool-call logs with JSON serialized into strings) keep the clone.

**Non-finite fast path (the JSON5 case itself).** The JSON5 fallback exists almost entirely because Python's `json.dumps` emits bare `NaN`/`Infinity`/`-Infinity` — files that are strict JSON except for a few tokens, paying a ~100× parser penalty for them. When `JSON.parse` rejects a payload, a single-pass in-string-aware scanner swaps bare non-finite tokens for nonce sentinels so native `JSON.parse` can take over (tokens in unquoted-key position route to `JSON5.parse` so key names survive). A restoring *reviver* proved ~8× slower than a plain parse, so the worker plain-parses the repaired text and walks it off-thread to locate the sentinels: small/string-heavy results get a targeted in-worker fixup and clone back; big node-dense ones ship the repaired text plus sentinel paths, and the main thread pays one plain `JSON.parse` plus a ~6ms targeted fixup. Real JSON5 syntax (comments, unquoted keys) still falls back to `JSON5.parse`.

**Robustness:**
- `NaN`/`Infinity` from JSON5 content survive large results intact (the old stringify round-trip silently turned them into `null`)
- A fatally-errored worker (e.g. OOM) is terminated and replaced in its pool slot instead of hanging every request routed to it; `onmessageerror` and pool `terminate()` reject in-flight requests; a `postMessage` throw (detached-buffer `DataCloneError`) can't leak an unsettleable promise
- Sentinel walks are iterative — `JSON.parse` accepts nesting far deeper than the JS call stack (covered by a 20k-depth test)
- Least-loaded worker dispatch, so small requests don't queue behind a multi-second parse while other workers idle

## Results (medians, `asyncJsonParse`, original → this PR)

| payload | total | main-thread blocked |
|---|---|---|
| scout events 186MB (real, node-dense) | 8.2s → **3.8s** | 3.9s → **1.9s** |
| synthetic eval log 100MB | 1312ms → **512ms** | 649ms → **165ms** |
| eval sample 31MB (real) | 540ms → **358ms** | 244ms → **109ms** |
| eval sample 18.5MB (real) | 280ms → **215ms** | 125ms → **46ms** |
| 50MB with bare `NaN` | 9.8s → **921ms** (11×) | 1.5s → **251ms** (6×) |
| 5MB with bare `NaN` | 750ms → **129ms** | — |
| summaries 58MB (real, string-heavy) | — | 124ms, max single stall 50ms |
| json5 5MB (real JSON5 syntax) | unchanged | 33–65ms (vs 674ms if parsed on main) |

For the 186MB dense file, ~1.9s blocked is essentially the floor: a bare `JSON.parse` costs 1.6s on the main thread and no transport can materialize a 36M-node graph for less. Going lower is an architectural question (streaming/pagination), out of scope here. A known follow-up: big dense payloads are parsed once in the worker (validation) and once on the main thread (materialization) in series; probing before parsing could cut scout's total ~40% at the cost of a resubmit path.

## Benchmarks (`packages/util/bench/`)

- Playwright/Chromium harness (real Web Workers) measuring total latency, main-thread blocked time (event-loop-gap sampling), and longest single stall, with correctness verification on every case — probes are non-finite-aware and compare exact non-finite value counts, so `NaN` → `null` corruption fails verification
- Deterministic synthetic payloads from 120B to 100MB across seven shapes (flat records, string-heavy eval-log-like, numbers, deep nesting, JSON5-only, strict-JSON-with-bare-`NaN`, JSON-serialized-into-strings)
- Real-payload support: drop `.json` files into `bench/fixtures/` (gitignored) and they run as cases; `--fixtures-only`, `--cases`, `--quick`, and `--compare a b` flags
- `pnpm -C packages/util bench -- --label <name>`; snapshots committed under `bench/results/`

## Testing

- Unit tests for the main-thread parse paths: JSON, JSON5 fallback, NaN/Infinity preservation (values, unquoted keys, root position), repair-scanner edge cases (tokens inside strings, escaped quotes, exponents, literals, malformed input), 20k-deep nesting, byte views, error cases; worker paths verified end-to-end by the harness's per-case correctness checks (108 case×API probes, all passing)
- Full workspace green: typecheck, lint, tests (662), format, manypkg

https://claude.ai/code/session_01XmudAyM3DweoAdEeX3ibaH


dragonstyle:
### Code review

Found 1 issue:

1. `repairNonFiniteJson` replaces bare `NaN`/`Infinity` tokens in object-**key** position, silently corrupting JSON5 input with unquoted non-finite keys. The scanner does not distinguish key from value position, so `{NaN: 1}` is repaired to `{"__json5_nan_<nonce>__": 1}` — valid strict JSON that parses successfully, and since the restore step (reviver / `findSentinelPaths`) only walks values, the sentinel key is never restored. Verified by running the PR-head code: `jsonParse('{NaN: 1}')` returns `{"__json5_nan_guxr2gfwys4__": 1}` (same for `{Infinity: 2}`), with no error signaled; before this PR, `JSON5.parse` returned `{ "NaN": 1 }`. This also contradicts the function's own doc comment ("Returns null if the text needs real JSON5 (comments, unquoted keys, quotes, ...)") — the `N`/`I` branches fire before the unknown-character bail-out is reached. Note the common Python case is unaffected (`json.dumps` quotes non-finite dict keys and only emits bare tokens in value position), so this only bites genuine JSON5 with unquoted `NaN`/`Infinity`/`-Infinity` keys — but the failure mode is silent data corruption. A fix could track whether the token is followed (after whitespace) by `:` while outside a string, and return `null` in that case, or verify after `JSON.parse` that the number of restored sentinels equals the number of replacements and fall back to `JSON5.parse` on mismatch.

https://github.com/meridianlabs-ai/ts-mono/blob/8509a717bf695900aa4748c400edf644f79a097a/packages/util/src/json-worker.ts#L199-L213

dragonstyle:
Confirmed and fixed in 2e86ce3 (the review was against the earlier head 8509a71). `repairNonFiniteJson` now checks whether a matched bare token is followed — after optional whitespace, outside strings — by `:`, and returns `null` in that case so the document routes to `JSON5.parse` and key names survive.

Verified against this comment's exact repro on the current head:

```
jsonParse('{NaN: 1}')          → {"NaN": 1}
jsonParse('{Infinity: 2}')     → {"Infinity": 2}
jsonParse('{"a": {NaN: true}}') → {"a": {"NaN": true}}
jsonParse('{NaN: 1, "x": NaN}') → key "NaN" preserved, value x is NaN
```

Unit tests cover key position (top-level, nested, whitespace-before-colon, and key+value mixed in one document) in `json-worker.test.ts` under "bare tokens in key position keep their names".

---

PR #389, opened 2026-07-08
viewer: AG Grid → TanStack DataGrid, react-query data-flow refactor, zustand devtools

brandly:
Groundwork for upcoming scaling work (paginated, database-backed listing queries / filtering / sorting). Three main changes:

**1. AG Grid → TanStack.** The log-list and samples grids move onto an in-house `DataGrid` built on TanStack Table + TanStack Virtual: virtualized rows, column resize/reorder/pinning with per-scope width persistence, keyboard nav + a11y, rotated score headers with heat-map scales. Filtering moves off AG Grid's `FilterModel` onto engine-agnostic `FilterSpec`s (`filterSpecRegistry`, `astToSpecs`/`specsToFilterText`, shared `columnFilter` package) with a SQL-semantics listing evaluator that stands in for a future server-side listing. AG Grid remains only in `ScoreAgGrid` (results scores table) plus a few lingering `ColDef` type imports.

**2. Tasks/samples data-flow refactor.** Server data leaves zustand (now UI state only) and moves to react-query behind two encapsulated subsystems: `app_config` (startup modes, API resolution, single-file / VS Code / dir gating) and `log_data` (`fetchEngine` — one prioritized fetch path for listing/details/samples/pending, listing sync, polling as engine producers). Imperative `*SideEffect` hooks become route-mounted controllers (`LoaderHost`, `LogLoadController`, `SampleLoadController`). Boundaries documented in `apps/inspect/design/domain-ownership.md`.

**3. Zustand devtools.** New `@tsmono/zustand-devtools` package (live state-tree panel), wired into the app alongside `@tanstack/react-devtools`.

**Also fixes bugs that exist on main** (surfaced by side-by-side validation against the origin/main viewer):

- A corrupt log in a directory poisoned the entire listing (every row "-", endless retry loop); now isolated per-file with bounded retries.
- `?log_dir` + `?log_file` together silently switched to static-http mode and died on `file://` fetches (blank page); now an explicit mutually-exclusive error.
- Opening a deleted log silently rendered stale cached samples; now an error page.
- A score named after a built-in filter variable (e.g. `epoch`) shadowed the built-in in filter expressions.
- `PopOver` painted at the wrong position for a few frames (visible jump) when its reference element sits in a CSS-transformed subtree.


apps/inspect/src/app/routing/loaders/SampleLoadController.tsx
```diff
@@ -0,0 +1,62 @@
...
+const kSampleBagKeys = ["scrollPosition", "listPosition"];
+
+/**
+ * Reacts to the selected sample changing — no fetching (the sample queries are
+ * mounted by the detail views through `useSelectedEvalSampleData`). Resets the per-sample
+ * UI state that isn't derivable from the new sample: scroll/list positions,
+ * collapsed events, and the timeline selection. Keyed on the sample's identity
+ * so re-selecting (or a running sample finalizing in place) doesn't reset.
+ */
+export const SampleLoadController: FC = () => {
```
brandly, line 17: This always returns `null` and could be a hook. It's only mounted in `LoaderHost.tsx` where it's mounted 100% of the time.

brandly: I guess the argument against changing it is that you'd lose the ability to conditionally render it. `SampleRouteSelectionController` also always returns null, but it's conditionally rendered.

apps/inspect/src/app/log-list/grid/LogListGrid.tsx
```diff
@@ -295,242 +214,275 @@ export const LogListGrid: FC<LogListGridProps> = ({
...
-      const cols = e.api.getColumnDefs();
-      if (cols && cols.length > maxColCount.current) {
-        maxColCount.current = cols.length;
-        refitColumns();
-      }
-    },
-    [refitColumns]
-  );
+    setFilteredCount(folders.length + total_count);
+  }, [folders.length, total_count, setFilteredCount]);
```
brandly: this number is getting shoved into Zustand really just so it can be rendered elsewhere. i'll see if it's easy to lift this value up instead

brandly: fixed in [971ee64](https://github.com/meridianlabs-ai/ts-mono/pull/389/commits/971ee64a51156fec68ede45d6951306a3574b799)

PR #404, opened 2026-07-10
Enable strict unnecessary condition linting

epatey:
Enable `@typescript-eslint/no-unnecessary-condition` (error) in the shared ESLint base config so new code gets the rule immediately.

Existing violations are baselined, not fixed:
- **`?.` collapses (107 sites)**: applied the rule&#39;s own suggestion fixes only where non-nullishness is guaranteed at runtime (a dominating guard in the same flow, locally constructed values, spec-guaranteed DOM behavior). Every one of the 179 candidate sites was audited.
- **`?.` kept (72 sites)**: where the guard is intentional defense — data isn&#39;t validated at the wire (old eval logs / journal files / API responses / persisted webview state may omit type-required fields), browser APIs that may be absent (`checkVisibility`, `activeElement`), or internal return types that lie via `@ts-expect-error`. Each carries a suppression plus an `intentional` comment saying why.
- **everything else suppressed**: file-level `eslint-disable` for files with &gt;3 violations, `eslint-disable-next-line` otherwise.

No behavior changes intended, with one deliberate exception: in `filters.ts` `sampleFilterItems`, the `!descriptor` fallback now runs before the `descriptor.filterable` read. Previously a missing score descriptor threw a TypeError before reaching the fallback (dead code); it now produces the placeholder item as intended, and `SampleFilterItem.scoreType` is typed `string | undefined` to match. lint / typecheck / test / format:check green across all packages.

### DIFF
--- a/apps/inspect/src/app/log-list/grid/columns/hooks.tsx
+++ b/apps/inspect/src/app/log-list/grid/columns/hooks.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unnecessary-condition */
 import clsx from "clsx";
 import { useCallback, useMemo } from "react";
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,6 +26,15 @@ TypeScript monorepo sharing code between inspect_ai, inspect_scout, vs code exte
   - A cast is a last resort for boundaries TypeScript can't express, with
     a comment saying why
 
+  **Parsed data: the types lie (#555).** Eval logs, journal files, API
+  responses, and persisted state are cast at the boundary, not validated —
+  old files omit fields the types declare required. Defensive `?.`/guards
+  on such data are intentional; do not remove them because the type (or
+  `no-unnecessary-condition`) says they're impossible. They carry
+  suppressions marked `intentional: data isn't validated at the wire
+  (#555)` and can only be removed by fixing issue #555 (validate at the
+  boundary).
+
 ## Code Style — Comments                                                       
                                                                                 
   Add comments only for non-obvious decisions:                                  
--- a/apps/inspect/src/app/log-list/LogsPanel.tsx
+++ b/apps/inspect/src/app/log-list/LogsPanel.tsx
@@ -81,6 +81,7 @@ export const LogsPanel: FC<LogsPanelProps> = ({
   // a different scope (different dir), so each folder also remembers its
   // own state. `undefined` until logDir hydrates so we never write under
   // a half-initialized scope.
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
   const scopeKey = logDir === undefined ? undefined : `${mode}::${currentDir}`;
 
   // Cache identity of the row universe: the listing/overview queries depend
--- a/apps/inspect/src/app/log-list/grid/logListRow.ts
+++ b/apps/inspect/src/app/log-list/grid/logListRow.ts
@@ -17,7 +17,7 @@ export const buildLogListRow = (item: LogListItem): LogListRow => {
   const derived = log?.derived;
 
   const taskArgsSource =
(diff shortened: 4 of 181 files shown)


apps/inspect/src/app/log-list/grid/columns/hooks.tsx
```diff
@@ -409,7 +409,7 @@ export const useLogListColumns = (
         meta: { sortComparator: numberCompare },
         accessorFn: (row) => row.totalSamples,
         cell: ({ getValue }) => {
-          const value = getValue<number | undefined>();
+          const value = getValue<number | null | undefined>();
           if (value === undefined || value === null) {
```
epatey: Was this code being paranoid checking for both undefined and null? Do we know that it's actually possible for null to get in there? It looks to me that getValue is typed to return unknown.

apps/inspect/src/app/log-list/grid/useLogListData.ts
```diff
@@ -52,6 +52,7 @@ const buildLogListRow = (item: LogListItem): LogListRow => {
   // actually supplied at the call site) over `task_args` (which
   // would also include defaulted values).
   const taskArgsSource =
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- eval is required in the generated header type but partial or older headers can omit it
```
epatey, line 55: This is a big deal because, if true it contradicts our general desire to trust TypeScript types. thoughts?

epatey:
Comments like this are a pretty big smell
> eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- config is required in the generated type but can be absent in logs from older writers

It's fine to be tolerant of old content that may not conform, but requiring the UI layer to have to deal with it is bogus. Let's think of a clean way to do data layer normalization so that the UI layer can truly trust the types. Of course, the approach will depend on how broad the set of "exceptions" is. Don't make any changes yet, but research and make a proposal.

epatey:
@auto the goal is to get the new lint rule enabled for new code that gets written. there's a huge backlog of violations. we can't let that stop us. we have to stop the bleeding. that said, we need to be paranoid because we do no validation at the boundary. this means that although we want to rely on our typescript types, external data often gets cast to the type with no actual checking. with that in mind, review the code with an eye toward introducing extremely little risk.

i-am-marvin:
Reviewing this PR with a focus on runtime risk from removed guards 

### Todo
- [ ] Gather context: full diff vs origin/main
- [ ] Audit every site where a `?.` or condition was *removed* (the only real risk surface)
- [ ] Verify low-risk categories (eslint-disable additions, config change)
- [ ] Post review findings

[View job run](https://github.com/meridianlabs-ai/ts-mono/actions/runs/32384184937)

epatey:
@auto use your best conservative judgement to address the review feedback.

i-am-marvin:
**Addressing review feedback conservatively** 

- [ ] Merge latest origin/main into this branch
- [ ] Fix `filters.ts`: move the `if (!descriptor)` defense above the `.filterable` deref (finding 1)
- [ ] Harden `AutocompleteInput.tsx` scroll effect with a bounds check (finding 2)
- [ ] Clean up backwards collapses (dead `??` after removed `?.`) in test files (finding 3)
- [ ] Restore defensive access in shared `NavPills.tsx` (finding 4)
- [ ] Remove stale suppression in `SampleSummaryView.tsx`
- [ ] Investigate + respond to review thread: `hooks.tsx` `null | undefined` question
- [ ] Respond to review thread: `useLogListData.ts` "types lie" concern
- [ ] Run lint / typecheck / tests
- [ ] Commit, push, request re-review

[View job run](https://github.com/meridianlabs-ai/ts-mono/actions/runs/32386733182)

i-am-marvin:
@review

PR #426, opened 2026-07-16
Share the query builder across Inspect and Scout

brandly:
## Summary

- replace the duplicate Scout query builder and hand-written query types with @tsmono/inspect-common/query
- keep the transcript-specific column registry in Scout with focused coverage
- verify the shared wire types against the generated Scout OpenAPI schema at compile time
- preserve the existing Scout query API through an explicit export boundary

## Why

Inspect and Scout had matching implementations of the same filter and ordering protocol. Making inspect-common/query the shared domain layer removes the duplicate implementation while keeping Scout-specific schema knowledge local to Scout.

## Verification

- @tsmono/inspect-common typecheck, lint, and tests: 59 passing
- Scout typecheck and lint
- focused Scout query tests: 6 passing
- Prettier and git diff checks


brandly:
Addressed both non-blocking review notes in 7b45f2bd:

- changed the module example to use a valid relative import
- made query/index.ts the single complete export boundary, migrated all remaining query/types imports, and removed the redundant shim

Verified with Scout typecheck, lint, formatting, and all 208 tests.

epatey (approved):

PR #428, opened 2026-07-16
React Doctor first pass

brandly:
[React Doctor](https://www.react.doctor/) produced a lot of feedback about our repo. It's mostly warnings, but it's still listing some things as bugs after these changes.

This PR is mostly low-hanging fruit. I ran `npx -y react-doctor@latest . -y` in the root of the repo and had Claude work on the results. Some of this is just cleaning up unused dependencies, usually dependencies that have been lifted up to a package in the repo, so it doesn't have much impact. In other cases, we weren't always handling `setTimeout`s well and had opportunities to leak memory.


packages/react/src/virtual/VirtualList.tsx
```diff
@@ -172,7 +172,11 @@ export function VirtualList<T>({
       const el = getScrollElement();
       if (el) {
         setFollowOutput(false);
-        setTimeout(() => el.scrollTo({ top: 0, behavior: "auto" }), 100);
+        const id = setTimeout(
+          () => el.scrollTo({ top: 0, behavior: "auto" }),
+          100
+        );
+        return () => clearTimeout(id);
```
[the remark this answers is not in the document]
brandly: @claude okay fix this please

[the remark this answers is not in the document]
i-am-marvin: **Claude finished @brandly's task in 2m 27s** —— [View job](https://github.com/meridianlabs-ai/ts-mono/actions/runs/29530053418)

---
### Review findings fixed ✅

- [x] Merge `origin/main` into the branch (clean merge, included in the push)
- [x] Fix `VirtualList.tsx` — the scroll-to-top timer id now lives in a ref cleared on unmount only, so `setFollowOutput(false)` re-running the effect no longer cancels it before it fires; `scrollToTopOnFinish` works again and the timer still can't leak past unmount
- [x] Fix `SampleDisplay.tsx` — hoisted the `focusOnLoad` check so no timer is scheduled when it's false
- [x] Commit and push (`790f4ff`)
- [x] Resolve the two addressed review threads

**Note:** I couldn't run the typecheck/lint locally because dependency install isn't permitted in this environment — both changes are small and use `useRef`/`useEffect` already imported in the files, but CI should confirm.
 · [Branch](https://github.com/meridianlabs-ai/ts-mono/tree/brandly/react-doctor-1)

apps/inspect/src/components/LargeModal.tsx
```diff
@@ -136,7 +136,7 @@ export const LargeModal: FC<LargeModalProps> = ({
               onClick={onHide}
               aria-label="Close"
             >
-              <HtmlEntity html={"&times;"} />
+              ×
```
epatey: Any insight into this change? I'm failing to parse it.

packages/react/src/hooks/useRafThrottle.ts
```diff
@@ -9,10 +9,12 @@ export function useRafThrottle<T extends (...args: unknown[]) => unknown>(
...
   const callbackRef = useRef<T>(callback);
 
-  // Update the callback ref when the callback changes
+  // Callers pass a fresh closure every render, so the latest-ref sync must run
+  // on every commit; a [callback] dep would just be a reference that never
+  // matches (and makes callers churn this effect on each render).
   useEffect(() => {
     callbackRef.current = callback;
-  }, [callback]);
+  });
```
epatey, line 17: This one's interesting. I'm under the impression that omitting the dependency array means execute the effect on/after every render. That's binding us to the behavior that occurs if the caller passes a different `callback` on every render. It just feels funny to do it this way. 1). It's way more subtle for the reader to omit the dependency array. 2). what if the callers changed their behavior? its odd coupling.

of course, I could be wrong. 

Also, wth is a raf. is it a typo'ed ref?

epatey (approved):
approved with minor questions

apps/inspect/src/components/LargeModal.tsx
brandly: It's just directly using the character instead of going through `dangerouslySetInnerHTML`. No real change in behavior but I guess it's nice not to have `dangerouslySetInnerHTML` around

packages/react/src/hooks/useRafThrottle.ts
brandly: Request Animation Frame! You're giving it a callback that only fires once per animation frame even if you call it multiple times in quick succession.

I think I'll switch this one back. I agree that it doesn't make sense to run it if `callback` doesn't change. (I'll poke Claude and see if it can explain itself first)

apps/inspect/src/components/LargeModal.tsx
epatey: oh `&times;` is `x`. lol. I see.

PR #431, opened 2026-07-17
fix(viewer): preserve narrow server timelines

jstar0:
## Summary

When server-authored timelines each cover a narrow subset of a sample, the
viewer currently merges unreferenced events into one arbitrary timeline. That
timeline then contains events outside its authored slice while its siblings
remain unchanged.

## Changes

- When two or more server timelines leave any events unreferenced, append an
  `Overall` timeline backed by the full built transcript and leave every
  authored timeline unchanged.
- Preserve the existing orphan merge for a single server timeline, where stray
  scorer events have one unambiguous host.
- Append rather than prepend `Overall`, so event and message deep links continue
  to select an authored timeline first when it contains the target.
- Clear the shared swimlane selection when switching timelines, preventing a
  row key from the previous timeline from partially matching the next one.
- Cover majority and minority orphan sets, single-timeline merging, no-orphan
  behavior, deep-link priority, and selection reset with regression tests.

## Verification

```console
pnpm --filter @tsmono/inspect-components test
pnpm --filter @tsmono/inspect-components typecheck
pnpm --filter @tsmono/inspect-components lint
pnpm --filter @tsmono/inspect-components format:check
pnpm build
```

The component suite passes with 867 tests and 2 existing skips.

Refs UKGovernmentBEIS/inspect_ai#4507


dragonstyle (changes requested):
please see comment

dragonstyle (changes requested):
Thanks for the careful follow-through here — this now tracks the agreed approach from UKGovernmentBEIS/inspect_ai#4507 exactly (count-based trigger with the minority-orphan regression locking it in, single-timeline merge preserved, `Overall` appended so authored timelines win deep-link resolution). I also traced the deep-link chain end-to-end (`useDeepLinkResolution` → `setActiveTimeline` → the Inspect `onSelect` adapter) and it behaves as intended for both programmatic and user-initiated switches.

One small regression to fix before merge, plus two optional nits — consolidated checklist:

- [ ] **Guard same-index switches in the new `setActiveTimeline` wrapper** (`useTranscriptTimeline.ts`). `TimelineSelector.handleSelect` calls `onSelect(index)` unconditionally, including for the already-active timeline. The wrapper then clears the swimlane selection, resets scroll to top, and strips `?event=`/`?message=` — where re-selecting the active timeline was previously a no-op. An early return when `index === activeTimelineIndex` fixes it (the programmatic path in `useDeepLinkResolution` already guards this, so only user clicks are affected). A small regression test would be welcome.
- [ ] *(optional)* The `if (options) { onSelectTimeline?.(null, options) } else { onSelectTimeline?.(null) }` branching can collapse to `onSelectTimeline?.(null, options)` — the adapter optional-chains `options`.
- [ ] *(optional)* A server-authored timeline literally named `Overall` would produce a duplicate dropdown entry. Unlikely in practice; harden only if you feel like it.

Once the guard is in, this is good to go. Thanks again for the patience through the redirections — the end state is much better than where we started.

jstar0:
Addressed the remaining same-index regression in 3104717d.

- Reselecting the active timeline now returns before clearing selection, resetting scroll, or stripping the deep link.
- Added a hook regression that verifies neither onSelect nor onActiveChange fires for a same-index selection.
- Cross-timeline selection reset and programmatic deep-link preservation remain unchanged.

Verification:
- pnpm --filter @tsmono/inspect-components test (906 passed, 2 skipped)
- pnpm --filter @tsmono/inspect-components typecheck
- pnpm --filter @tsmono/inspect-components lint
- pnpm --filter @tsmono/inspect-components format:check
- pnpm build

Please re-review when convenient.

dragonstyle (approved):

jstar0:
CI follow-up on the current  baseline: the generated  schema now includes optional , but Scout's exhaustive transcript column registry did not yet include that key, so  failed on both  and this PR.\n\nI pushed  to register the label, tooltip, opt-in column, and filter/order metadata. It is intentionally excluded from the default-visible set, so the existing table layout is unchanged.\n\nVerification on the exact head:\n- [ERR_PNPM_NO_IMPORTER_MANIFEST_FOUND] No package.json (or package.yaml, or package.json5) was found in "/Volumes/JSData/ComputerScience/AutoHunter". (including the suppressions ledger)\n- [ERR_PNPM_NO_IMPORTER_MANIFEST_FOUND] No package.json (or package.yaml, or package.json5) was found in "/Volumes/JSData/ComputerScience/AutoHunter".\n- [ERR_PNPM_JSON_PARSE] Expected property name or '}' in JSON at position 293 (line 13 column 5) in /Volumes/JSData/ComputerScience/AutoHunter/external-repos/agentcore-cli/src/assets/typescript/http/strands/base/package.json

  11 |   },
  12 |   "dependencies": {
> 13 |     {{#if (eq modelProvider "Anthropic")}}
     |     ^
  14 |     "@anthropic-ai/sdk": "^0.71.2",
  15 |     {{/if}}
  16 |     {{#if (eq modelProvider "OpenAI")}}: 974 passed, 2 skipped\n- Scout transcript-column tests: 21 passed\n- SSH commit signature verified\n\nHosted CI has restarted for ; please re-review when convenient.

jstar0:
CI follow-up on the current `main` baseline: the generated `TranscriptInfo` schema now includes optional `score_explanation`, but Scout's exhaustive transcript column registry did not yet include that key, so `typecheck` failed on both `main` and this PR.

I pushed `bc52a1d1` to register the label, tooltip, opt-in column, and filter/order metadata. It is intentionally excluded from the default-visible set, so the existing table layout is unchanged.

Verification on the exact head:
- `pnpm check` (including the suppressions ledger)
- `pnpm build`
- `pnpm --filter @tsmono/inspect-components test`: 974 passed, 2 skipped
- Scout transcript-column tests: 21 passed
- SSH commit signature verified

Hosted CI has restarted for `bc52a1d1`; please re-review when convenient.

PR #465, opened 2026-07-28
Don't blank the open log when an eval finishes (VS Code webview)

Aprillion:
Simplify vscode extension's `backgroundUpdate` (reportedly broken since October 2025 and not present in browser since February 2025) - when a new eval appeared in the same folder as currently displayed eval, it tried to toggle to the new one and failed:

<details>
<summary>Full report from Opus 5 including git archeology</summary>

## A finished eval blanked the open log

When the editor announced a log created or changed in the viewed directory and the webview was unfocused, the handler selected that log — switching the view away from whatever was open, and to a selection that could not load: the bare announced path was mis-joined onto the file:// log-directory URI, so every fetch 404'd, leaving a blank panel under the previous log's breadcrumb. The fix removes selection rather than repairing the path: every announcement, from any directory, now just refreshes the listing — which even the matching-directory case never did before.

Repro:
1. In the editor, open Inspect View, click into a log and then a sample, so you are on a sub-route rather than the listing.
2. Click into the editor so the webview loses focus — the old handler only fired while unfocused.
3. Leaving it unfocused, run an eval writing a new log into the directory the viewer is watching, e.g. inspect eval <task> --log-dir ./logs/<subdir>. A mock model is fine.

Before:
* Requests with the log directory in the path twice, each 404, retried. The sample goes blank while the breadcrumb still names the log you opened; the malformed identifier persists in webview storage, so the failures continue until storage is cleared.

After:
* Nothing user-visible: no request is issued and the sample you were reading is untouched. The new log appears in the listing next time you look at the folder.

Also fixed — reload restores your place:
* Reloading the editor window now reopens the last open log instead of coming up empty. Plausibly the same bug — startup seeds the selection, then an unfocused announcement overwrote it with an unloadable path — but observed rather than traced.

One behaviour goes with it. A finishing eval no longer opens itself even when nothing is open and you are sitting on the listing. That path was equally broken before, so nothing that worked is lost, but it is the one part of this change that is not purely a bug fix.

The diff's second file only adds a comment marking open_log_file as hawk's: it has no callers in this repo, but hawk both implements LogViewAPI and calls that member, so the tempting cleanup breaks hawk's typecheck. Deleting it was tried and reverted.

## How it got this way

The message and the handler arrived together in [d805800](https://github.com/UKGovernmentBEIS/inspect_ai/commit/d8058008382097666534056ba531a5f73bf8082e) (2024-09-04, "Improve loading performance"), shaped a day later by "Improve Task Complete Workflow". The intent was coherent: an eval finishes, you are not looking at the viewer, so it shows you the result.

The extension sent a bare path for the file and a URI for the directory from the very first commit — and still does. That was verified against the shipped bundle, which ships an unminified tree: the current release builds the payload identically to 2024. The extension never changed. The asymmetry was harmless by design, because the viewer suffix-matched the announced path against relative listing names (logUrl.endsWith(val.name)) and compared directories only for equality. Nothing was ever joined.

What broke it was [0ed6ad4](https://github.com/UKGovernmentBEIS/inspect_ai/commit/0ed6ad4962174bd64ab3e20e6a5042fff5cefac6) (2025-10-24, "Use absolute paths in tracking log selection (fix 403s)"), which replaced the suffix match with absolutisation — isUri(f) ? f : join(f, logDir). A bare path is not a URI, so it takes the join; and join's only guard returns early when the file already starts with the directory, which a bare path does not. Hence the directory twice. The background-update path is not mentioned in that commit: the break was collateral, and the same value reaches updateState safely because that path normalises with isUri() first.

The browser had the same feature, and lost it silently. It polled for events every second and, on a refresh-evals signal, reloaded the listing and selected index 0 — the newest log, since listings sort newest-first. That was disabled in [7463f0b](https://github.com/UKGovernmentBEIS/inspect_ai/commit/7463f0b5de03a9c5eaed50141680106488faf1d7) (2025-02-03, "Convert Viewer to Typescript / React") by passing pollForLogs={false} at the only entry point, with no mention in the commit message, and the code was removed a few weeks later on the grounds that the flag "is always set to false anyhow". So no commit anywhere records a decision that auto-selection was unwanted — it was dropped in a rewrite and then tidied away.

The rest of that plumbing is still live: the eval run still notifies, the server still reports refresh-evals, and the viewer still polls it — and does nothing with it but invalidate the listing. Which is precisely what this change makes the editor path do, so the two surfaces agree again for the first time since early 2025.

One consequence worth noting beyond this fix: a recent change made the browser's own log directory a canonical URI too, so the same join would double any bare path fed to log selection there. Worth a look if other callers pass unnormalised input.

## Not addressed

An identifier corrupted by an earlier version persists in the webview's storage and keeps being retried, so anyone who already hit this will see the failed fetches until they clear storage. This stops new corruption; it does not repair what is already stored.
</details>

Screenshot 2026-07-28 083019


dragonstyle (approved):
Thanks — traced this on both sides and it holds up. The extension posts `log_file.path` (bare path) with `log_dir.toString()` (URI), the embedded viewer's `logDir` is URI-form, so the same-directory branch always went through `join()` and produced the doubled path. The other branch (`open_log_file` for a different directory) is refused by the extension's scope check anyway. So nothing that currently works is lost, and the unfocused case now matches what the focused case and the browser viewer already do.

Approving. Take-or-leave:
- [ ] A small unit test that a `backgroundUpdate` message leaves the selected log untouched would stop someone re-adding the selection later.
- [ ] The `used by hawk` comment could instead note that `open_log_file` is part of the host API contract (every adapter here implements it), so the reason it stays doesn't depend on an external repo.

Plain-language summary: In the VS Code viewer, an eval finishing while you were looking elsewhere no longer replaces the log you had open with a blank panel; the log list refreshes instead.

PR #473, opened 2026-07-29
feat(inspect): let embedders render their own chrome around <App/>

rasmusfaber:
## Summary

An app that hosts the viewer in-process and renders its own chrome around `<App/>` — a status bar, a side pane keyed to the selected sample — has to read the viewer's selection state from outside `<App/>`. Since #389 it can't: the selection hooks (`useSelectedSampleSummary`, `useSelectedScores`, `useLogSelection`) resolve through react-query and the gated app config, so calling them from a sibling of `<App/>` throws `No QueryClient set`, and then `App config not loaded` once a provider is added — the hooks keep throwing until the config query settles. Neither piece is reachable from outside the package: `queryClient` is a private module singleton, and nothing reports when the config has resolved.

Export both, so an embedder can wrap its chrome in the viewer's own react-query client and hold that chrome off until the hooks are safe to call. Standalone `inspect view` is unaffected — `<App/>` already provides both for its own subtree.

## Usage

The provider goes above both, so the chrome and the viewer share one client:

```tsx
<InspectQueryClientProvider>
  <MyChrome />   {/* renders its hook-using parts only once useViewerReady() */}
  <App />
</InspectQueryClientProvider>
```

### DIFF
--- a/apps/inspect/src/app_config/index.ts
+++ b/apps/inspect/src/app_config/index.ts
@@ -18,6 +18,7 @@ export {
   APP_CONFIG_KEY, // TODO: Exported for tests?! review
   getApi,
   useAppConfig,
+  useAppConfigAsync,
   useAbsLogDir,
   useLogDir,
 } from "./hooks";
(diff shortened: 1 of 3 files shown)


brandly:
@rasmusfaber thanks for this! Sorry it fell through the cracks for a couple weeks!

PR #478, opened 2026-07-30
Messages data interface

brandly:
The goal of this PR is to put cleaner interfaces around our messages. The raw messages from the eval file are accessed like this:

```tsx
export interface SampleConversation {
  /** Exact length of the final conversation. Cheap for every format:
   *  inline arrays know their length, chunked shells carry ref widths. */
  readonly messageCount: number;
  /** Messages `[start, end)` of the conversation, fully resolved.
   *  Out-of-range bounds clamp to `[0, messageCount)`. */
  getMessages(start: number, end: number): Promise<ChatMessage[]>;
}
```

But some of these `ChatMessage` get folded together into a single `MessageRow`, which means we have two different coordinate systems. The UI is ultimately going to access `MessageRow`s page by page, and we'll do it through this interface:

```tsx
export interface SampleMessagesData {
  /** Exact folded-row count for the whole conversation. */
  rowCount(): Promise<number>;
  /** One page of resolved rows. */
  getRows(pagination: Pagination): Promise<MessageRowsPage>;
  /** The whole conversation as text (copy/download) — the only sanctioned
   *  full materialization. */
  exportText(): Promise<string>;
}

export interface MessageRowsPage {
  rows: MessageRow[];
  /** Absolute position of `rows[0]` in the folded row space. */
  offset: number;
  /** Count of ALL rows, not just this page. */
  totalRowCount: number;
  /** Cursor for the page after this one (null at the end). */
  nextCursor: Cursor | null;
  /** Forward-read anchor of the preceding page (null at the start). The
   *  window it opens may overlap this page's first rows when fewer than
   *  `limit` rows precede it. */
  prevCursor: Cursor | null;
}
```

This data is shaped such that we can paginate these rows, but at the moment, we still materialize all of the rows at once and huge samples are still slow. This sets us up to make further optimizations and ultimately to handle huge samples in an efficient manner.

These details are now removed from UI components. UI calls `useSampleMessages` and renders the rows its given -- whether or not the eval is chunked or if it's a live sample.

The naming should improve and I think @epatey has feedback on that front.


apps/inspect/src/app/samples/SampleDisplay.tsx
```diff
@@ -546,16 +545,29 @@ export const SampleDisplay: FC<SampleDisplayProps> = ({
...
+        // streaming samples have none, and a silent no-op menu item reads
+        // as broken (chunked samples hydrate on demand inside the export)
+        ...(exportMessages
+          ? {
+              Messages: () => {
+                // the confirm icon must wait for the clipboard write — it
+                // can reject (unfocused document), and flipping early
+                // reads as a false success
+                exportMessages()
+                  .then((text) => navigator.clipboard.writeText(text))
```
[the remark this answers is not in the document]
brandly: I'm okay with this for now. This `exportMessages` interface is likely to change, and we'll refine this behavior. Un-chunked evals will already be in memory, and the promise will resolve immediately.

packages/inspect-components/src/chat/ChatViewVirtualList.tsx
```diff
@@ -252,8 +249,32 @@ export const ChatViewVirtualList: FC<ChatViewVirtualListProps> = memo(
...
+      () =>
+        buildMessageRows(
+          messages,
+          messageRowOptions({ callStyle, collapseToolMessages })
+        ),
+      [messages, callStyle, collapseToolMessages]
+    );
+    return <ChatViewRowsVirtualList rows={rows} tools={tools} {...rest} />;
+  }
+);
```
brandly: I think this only exists to maintain the public interface of this package, but we're not using this anymore. We should probably get rid of this component and rename the file to `ChatViewRowsVirtualList.tsx`

apps/inspect/src/app/samples/SampleDisplay.tsx
```diff
@@ -587,15 +599,23 @@ export const SampleDisplay: FC<SampleDisplayProps> = ({
...
+          // offered only when a settled conversation exists to export (see
+          // the copy dropdown)
+          ...(exportMessages
+            ? {
+                Messages: () => {
+                  exportMessages()
+                    .then((text) =>
+                      text.length > 0
+                        ? api.download_file(`${sampleId}-messages.txt`, text)
+                        : undefined
```
brandly: not sure why we'd need to guard against empty strings. can probably remove this

packages/inspect-components/src/chat/ChatViewVirtualList.tsx
brandly: i was wrong here. `ChatViewVirtualList` is used in some other places

epatey (approved):
🚀

PR #500, opened 2026-08-07
Enable React Compiler in inspect and scout

brandly:
Wires `babel-plugin-react-compiler` into both apps via `@vitejs/plugin-react`'s `reactCompilerPreset`, and fixes one real bug the compiler exposed.

Opening this for a decision on whether the build-time cost is worth it — the correctness work is done and verified either way. See **Cost** below for the number, and **If the cost isn't acceptable** for the cheaper landing.

## How it's wired

`plugin-react` v6 dropped Babel entirely (its only runtime dep is `@rolldown/pluginutils`; rolldown/oxc does JSX and Fast Refresh natively). The compiler is a Babel transform, so it needs Babel added back explicitly — `babel-plugin-react-compiler`, `@rolldown/plugin-babel`, `@babel/core`, `@types/babel__core`, all declared as optional peers of `plugin-react`:

```ts
plugins: [
  react({ jsxRuntime: "automatic" }),
  babel({ presets: [reactCompilerPreset()] }),
]
```

Because the shared packages are consumed as TS source (`"exports": "./src/index.ts"`), enabling this per-app also compiles `@tsmono/react`, `inspect-components` and `scout-components` through that app's pipeline. No per-package config needed.

React 19.2 provides `react/compiler-runtime`, so no runtime package and no `target` option. The lib build externalizes it correctly via the existing `/^(react|react-dom)(\/|$)/` rule — one `import { c as _c } from "react/compiler-runtime"` in `lib/index.js`.

## The bug this surfaced

Scout would not boot with the compiler on — 63 of 64 e2e tests failed, in dev **and** in the production build. Bisected to `AppRouter.tsx` + `state/store.ts` compiled together.

Scout builds its zustand store per-instance and injects it through context, so `useStore` pulled the bound store out under the name `store`. React Compiler identifies hook calls by callee name (`use[A-Z]…`), so `store(selector)` looked like an ordinary call and got memoized:

```js
if ($[0] !== store || $[1] !== t0) { t1 = store(t0); … } else { t1 = $[2]; }
```

On the second render with an unchanged store and selector, that call doesn't execute — zustand's `useSyncExternalStore` never registers, every subsequent hook in the fiber shifts a slot, and React's `areHookInputsEqual` reads `deps` off a hook state object that has none: `Cannot read properties of undefined (reading 'length')`.

Renaming the local to `useBoundStore` makes it recognizable, and the compiler switches from silently miscompiling to safely declining:

```
CompileError | Hooks must be the same function on every render, but this value
             | may change over time to a different function.
```

Also collapsed the two `store(...)` call sites into one (branching around two separate calls was itself a conditional hook call) and hoisted the whole-state selector to module scope, which stops allocating a fresh selector every render.

`useStore` losing memoization costs nothing measurable — it's a context read plus a subscription. Its callers still compile.

**Worth knowing:** no static check caught this. Lint was green on that file, `react-compiler-healthcheck` reported 711/711 components compiled with no incompatible libraries, and running the compiler directly logged `CompileSuccess`. Only the e2e suite caught it. Treat "e2e green on both apps" as the gate here, not lint.

## Verification

Rebased onto `b0f4d04` and re-verified from scratch, since main added a Timeline tab, pagination and the messages interface that the compiler now transforms:

- inspect e2e **96/96**, scout e2e **64/64**
- `turbo run lint typecheck format:check test` — **36/36 tasks**
- both app builds and the publishable lib build succeed
- scanned the 205 source files main added for the same dynamic-hook-callee pattern — none found

## Cost

| | before | after |
|---|---|---|
| inspect `build` | 7.1s | 24.1s |
| scout `build` | 5.4s | 29.8s |
| dev cold-transform, 145 tsx | 3.3s | 29.1s |
| scout bundle (minified) | 2,926 KB | 3,160 KB (+8.0%) |
| inspect bundle (gzip) | 1,246 KB | 1,362 KB (+9.3%) |

Babel is 46–51% of plugin time. This hit is steeper than most React Compiler adopters see, because moving to Vite 8 had removed Babel from the pipeline entirely — Next.js users and anyone on `plugin-react` v4/v5 were already paying for a Babel pass. The dev figure is a deliberate worst case (forcing every `.tsx` through at once); what a developer actually feels is the ~200ms per-file HMR re-transform. In CI the `build` job gains ~41s across both apps, against a check suite that already runs ~3m.

Output: 712 functions compiled, 13,608 memo cache slots, 88 bail-outs across 821 files. Bail-outs are the compiler safely declining, not errors.

## If the cost isn't acceptable

`compilationMode: "annotation"` is nearly free — **8.8s** for inspect vs 7.1s baseline — because the preset filters at the rolldown layer before Babel runs, so files without a `"use memo"` directive never reach it. That's a one-line change, and it's also React's documented incremental path: opt in per file, flip to `infer` later. Given the one real defect here was invisible to every static check, staging it that way makes rollout bisectable.

## Not included

Deliberately left for separate decisions:

- `apps/inspect/src/state/store.ts:51` has the same hazard (`storeImplementation(selector)`), inert only because the `as UseBoundStore<…>` cast hides `useStore` from the compiler. Any refactor dropping that cast re-arms it.
- 63 `eslint-disable react-hooks/*` comments cause the compiler to skip 16 whole components, including `PopOver` and both `DataGrid`s.
- Neither app uses `StrictMode`; both `main.tsx` files call bare `createRoot`.
- Vitest doesn't run the compiler (no plugin-react in either `vitest.config.ts`), so unit tests exercise uncompiled code. Same for Storybook in `packages/react`.

---
_Generated by [Claude Code](https://claude.ai/code/session_011VaFgTNky8xT6xJBJg7ZZ4)_

### DIFF
--- a/apps/inspect/package.json
+++ b/apps/inspect/package.json
@@ -48,9 +48,11 @@
     "./styles/index.css": "./lib/styles/index.css"
   },
   "devDependencies": {
+    "@babel/core": "^7.29.7",
     "@eslint/js": "^10.0.1",
     "@msw/playwright": "^0.6.7",
     "@playwright/test": "^1.62.0",
+    "@rolldown/plugin-babel": "^0.2.3",
     "@tanstack/react-devtools": "^0.10.9",
     "@tanstack/react-query-devtools": "^5.101.4",
     "@testing-library/dom": "^10.4.1",
@@ -65,6 +67,7 @@
     "@tsmono/tsconfig": "workspace:*",
     "@tsmono/util": "workspace:*",
     "@tsmono/zustand-devtools": "workspace:*",
+    "@types/babel__core": "^7.20.5",
     "@types/bootstrap": "^5.2.11",
     "@types/clipboard": "^2.0.10",
     "@types/codemirror": "^5.60.15",
@@ -76,6 +79,7 @@
     "@types/react-dom": "^19.2.3",
     "@typescript/native": "npm:typescript@^7.0.2",
     "@vitejs/plugin-react": "^6.0.4",
+    "babel-plugin-react-compiler": "^1.0.0",
     "cross-env": "^10.1.0",
     "eslint": "^10.7.0",
     "eslint-plugin-react-hooks": "^7.1.1",
--- a/apps/inspect/vite.config.ts
+++ b/apps/inspect/vite.config.ts
@@ -1,7 +1,8 @@
 import { cpSync, rmSync } from "fs";
 import { join, resolve } from "path";
 
-import react from "@vitejs/plugin-react";
+import babel from "@rolldown/plugin-babel";
+import react, { reactCompilerPreset } from "@vitejs/plugin-react";
 import pc from "picocolors";
 import type { Plugin } from "vite";
 import { defineConfig } from "vite";
@@ -43,6 +44,7 @@ export default defineConfig(({ mode }) => {
         jsxRuntime: "automatic",
         fastRefresh: !isLibrary,
       }),
+      babel({ presets: [reactCompilerPreset()] }),
     ],
     resolve: {
       dedupe: [
--- a/apps/scout/package.json
+++ b/apps/scout/package.json
@@ -53,8 +53,10 @@
     "react-dom": "^19.0.0"
   },
   "devDependencies": {
+    "@babel/core": "^7.29.7",
     "@msw/playwright": "^0.6.7",
     "@playwright/test": "^1.62.0",
+    "@rolldown/plugin-babel": "^0.2.3",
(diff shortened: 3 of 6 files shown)


brandly:
It's still experimental but React Compiler was rewritten in Rust: https://github.com/react/react/pull/36173

My hope is that it lands soon, and our dev servers start up faster. In the meantime, this puts us in a position to stop doing so much `useMemo`/`useCallback` everywhere. It's a tradeoff in dev experience today, but the downside should go away soon.

brandly:
closing this in favor of #536. that one will merge once we wait out the soak window, which should be on Friday.

PR #505, opened 2026-08-08
Show images embedded in Markdown again (#502 + review fixes)

dragonstyle:
## Summary

This is @rasmusfaber's #502 with the two non-blocking review fixes applied, re-homed on a branch in this repo so it can be pushed to and merged directly (the #502 head lives on a fork that the review bot cannot push to). **All substantive work here is Rasmus Faber-Espensen's — his 10 commits are cherry-picked with authorship intact.** The one additional commit is the review-fix commit from `marvin/pr-502-image-fixes` (also cherry-picked as-is), which addresses the two inline review comments on #502:

- the markdown image rule now emits `canonicalImageSource(source)` rather than the raw source, so the documented gate-then-render-canonical invariant holds locally in `renderMarkdown` output;
- `MarkdownDiv.css` caps markdown images at `max-width: min(100%, 800px)`, matching the typed `ContentImage` path and preventing horizontal overflow of the transcript column.

Supersedes #502 — if this merges, #502 should be closed with a pointer here.

## Original description (from #502)

Markdown stopped displaying embedded images in #362. That PR replaced the markdown-it image rule with a link/text fallback and added `img` to the sanitizer's forbidden tags, to stop transcript content from auto-loading remote media: loading a remote image leaks a fetch to an attacker-controlled host. But it also caught base64 data URIs, which issue no network request at all. The typed-content path still renders those, so identical bytes display or don't purely by whether they arrive as a `ContentImage` part or inside Markdown.

Reported case, from an Info event, which rendered as text: `noise_10_step (data:image/gif;base64,...)`.

This makes the Markdown path consult the predicate the typed path already uses, `isRenderableImageSource`, so base64 raster data URIs render as ``. Remote Markdown images are unchanged: still links.

### Notes for reviewers

The remote-image guarantee moved. `img` is no longer in the sanitizer's `FORBID_TAGS`; instead every `img` `src` is gated on `isRenderableImageSource`, and the element is removed when its src is rejected. Since markdown-it runs with `html: true`, this is also what now blocks literal `` in transcript content.

The result is strictly narrower than pre-#362, which allowed any `data:image/*` on `img` including `image/svg+xml`. SVG stays blocked here regardless of encoding, and non-base64 data URIs are rejected too.

### Known parity gap

Markdown image syntax never reaches the image rule unless markdown-it's own `validateLink` accepts the destination, and its built-in check only passes `data:image/(gif|png|jpeg|webp)`. So base64 data URIs for avif, bmp, x-icon / vnd.microsoft.icon — and the `image/jpg` alias — still render as literal text in Markdown even though the typed `ContentImage` path renders them. Overriding `validateLink` would widen link handling generally, so this PR leaves the gap in place.

## Verification

- `pnpm check`: 27/27 tasks pass
- `pnpm --filter @tsmono/react --filter @tsmono/util test`: all pass (react: 211 tests / 27 files)
- Branch tree verified byte-identical to `marvin/pr-502-image-fixes` after cherry-picking onto current `main`

🤖 Branch assembled with [Claude Code](https://claude.com/claude-code); all feature work authored by @rasmusfaber.


packages/react/src/components/renderedHtmlSanitizer.ts
```diff
@@ -212,6 +244,23 @@ const sanitizeStyleAttributeHook = (
...
+  node.tagName.toLowerCase() === "img";
+
+/**
+ * A remote src is not unsafe by protocol, but fetching it leaks a request to an
+ * attacker-chosen host, so an img src is limited to inline raster data.
+ *
+ * Returns the value to keep, so the attribute that lands in the DOM is the one
+ * that was validated.
+ */
+const safeImgSrc = (value: string): string | undefined => {
```
dragonstyle, line 257: **Pre-existing gap, suggest a follow-up rather than a change here.** The PR's premise — transcript content must not trigger remote fetches — has one legacy channel left open that predates this change: the `background` attribute. It's in `URL_ATTRIBUTES`, but `isSafeUrlAttribute` only rejects `data:`/`javascript:`/`vbscript:`, so an http(s) value passes, and DOMPurify's default profile keeps the attribute. Verified empirically against this project's DOMPurify:

```
sanitize('<table background="https://evil.example/x.png"><tr><td>x</td></tr></table>')
→ <table background="https://evil.example/x.png">…
```

Chrome and Firefox still honor `background` on `table`/`body` as a background image, so this fetches from an attacker-chosen host. Not introduced by this PR (`img` was forbidden while `background` never was), but now that the remote-image guarantee lives in `safeImgSrc`, adding `"background"` to `FORBID_ATTR` would close the last obvious image-fetch channel — nothing legitimate in transcript content uses it.

dragonstyle (commented):
## Code review (Claude Code, fresh context)

### Scope
Full review of the combined change (rasmus's #502 plus the review-fix commit) against `main`, with the sanitizer read in full context and one attack-surface question verified empirically.

### Verification performed
- Branch tree confirmed byte-identical to `marvin/pr-502-image-fixes` after cherry-pick onto current `main`; all 10 feature commits retain @rasmusfaber's authorship
- `pnpm check`: 27/27 tasks pass; `@tsmono/util` + `@tsmono/react` test suites pass (211 react tests)
- Confirmed both #502 inline review comments are addressed by the fix commit: the image rule emits `canonicalImageSource(source)` through `escapeHtml`, and `MarkdownDiv.css` caps images at `min(100%, 800px)`
- Probed DOMPurify defaults directly (jsdom) for adjacent fetch channels

### Assessment — no blocking issues
The security design holds up well under adversarial reading. `img src` really is the only newly opened channel: `srcset`/`srcdoc` were already in `FORBID_ATTR`, and `picture`, `source`, `video`, `audio`, `track`, `image`, `input` were already forbidden tags, so the classic alternate image-fetch routes are closed. The gate is enforced twice (`uponSanitizeAttribute` canonicalizes-or-strips; `afterSanitizeAttributes` removes any img that still lacks a valid src), which also kills the broken-image-placeholder case. The canonicalization invariant — validate the exact string you emit — is now held locally in both the markdown rule and the sanitizer, and the test for U+FEFF-in-scheme/path-traversal explains precisely the failure mode it pins. Test coverage is exemplary: unit (predicate + canonicalizer), sanitizer-level (14 rejection cases), pipeline-level (escaping driven through the markdown-it instance to avoid masking), and e2e in both apps.

One pre-existing gap found (inline comment below): the legacy `background` attribute is a remote-fetch channel that survives sanitization. It predates this PR — `img` was forbidden but `background` never was — so it's a follow-up, not a blocker here.

One observation, not worth changing: `canonicalImageSource` parses the URL twice (once inside `parseDataUri`, once for `.href`); negligible for strings this size.

packages/react/src/components/renderedHtmlSanitizer.ts
dragonstyle: Added a tracking item internally

packages/react/src/components/markdownRendering.ts
```diff
@@ -79,14 +83,23 @@ export const getMarkdownInstance = async (
...
 
     const source = token.attrGet("src") ?? "";
+    const alt = token.content.trim();
+
+    // Base64 raster data URIs issue no network request, so rendering them
+    // inline cannot leak a fetch to an attacker-controlled host. Emit the
+    // canonical form, not the raw source — see canonicalImageSource.
+    const canonicalSource = canonicalImageSource(source);
+    if (canonicalSource !== undefined) {
+      return `<img src="${md.utils.escapeHtml(canonicalSource)}" alt="${md.utils.escapeHtml(alt)}">`;
```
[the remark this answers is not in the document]
i-am-marvin: Fixed in af140323 — the image rule now reads `token.attrGet("title")` and emits it escaped the same way as `alt`, with rule-level tests (emitted / escaped / omitted when absent) and a pipeline-level test that the title survives sanitization.

PR #507, opened 2026-08-10
fix: resolve attachment refs in scan results

rasmusfaber:
## Summary

Scan results render `attachment://<hash>` as literal text wherever the recorded transcript kept its attachment refs unresolved. inspect_ai turns every text value over 100 characters into an attachment, so on a real transcript that is most of the content: system prompts, model outputs, tool results.

The results path already expands pooled events but never resolved attachments, because until now the recorded `input` arrived pre-resolved. The transcripts path has done this all along (`api-scout-server.ts` calls `resolveAttachments` after `expandEvents`); this applies the same existing helper to the scan-results path, reading the table from `input_data.attachments`.

Needed by meridianlabs-ai/inspect_scout#491, which records the compact form rather than inlining every duplicate.

### DIFF
--- a/apps/scout/src/api/expandInputEvents.test.ts
+++ b/apps/scout/src/api/expandInputEvents.test.ts
@@ -1,0 +1,23 @@
+import { describe, expect, it } from "vitest";
+
+import { expandInputEvents } from "./expandInputEvents";
+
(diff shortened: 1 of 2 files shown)


brandly:
superseded by #550

PR #527, opened 2026-08-14
Prevent WorkQueue retries after clear

### LINKED ISSUE #405: WorkQueue.clear can resurrect failed in-flight work in a new engine session
## Summary

WorkQueue.clear removes only items that are still present in itemsById. A batch already claimed by a worker is held in a local variable and is unaffected by clear. If that in-flight batch fails after clear, the retry branch reinserts it into itemsById.

In the fetch engine, stop calls queue.clear and advances the engine epoch during log-directory changes. A failed old-directory request can therefore be reinserted after stop, reclaimed under the new epoch, and executed using the new directory dependencies.

Introduced/exposed by the fetch-engine and queue refactor in 16a17f46.

Relevant code:

- https://github.com/meridianlabs-ai/ts-mono/blob/16a17f46/apps/inspect/src/utils/workQueue.ts#L112-L154
- https://github.com/meridianlabs-ai/ts-mono/blob/16a17f46/apps/inspect/src/utils/workQueue.ts#L199-L201
- https://github.com/meridianlabs-ai/ts-mono/blob/16a17f46/apps/inspect/src/log_data/fetchEngine.ts#L656-L676

## Failure sequence

1. A fetch for directory A is claimed and removed from itemsById.
2. FetchEngine.stop runs during a switch to directory B and calls queue.clear.
3. The A request fails after stop.
4. WorkQueue.runWorker increments its retry count and reinserts the claimed item.
5. The worker claims it again after the B engine is active.
6. FetchEngine stamps the retry with the current B epoch and uses the current B dependencies.

Because the retry receives a fresh epoch when reclaimed, FetchEngine.dropStaleEpoch no longer recognizes it as A work. An A filename can be requested from and written through the B session.

## Impact

A transient fetch failure during a root switch can bypass the engine epoch fence and mix work between directories. The same WorkQueue behavior also makes clear semantically weaker than its API suggests for any future consumer.

## Suggested direction

Give WorkQueue a generation or cancellation token:

- Increment the generation in clear.
- Capture it when a batch is claimed.
- Never retry or complete a batch whose captured generation no longer matches.
- Consider exposing cancellation/settlement semantics explicitly if clear must reject or report abandoned work.

The fetch engine should still retain its own epoch checks as defense in depth.

## Acceptance criteria

- A deterministic test claims an item, calls clear while the worker is paused, then releases the worker with a retryable failure.
- The item is not reinserted or executed again.
- A fetch-engine regression test switches from A to B during a failing A request and verifies no A request or sink write occurs under B dependencies.

fraware:
Fixes #405.

`clear()` now invalidates claimed batches so stale work cannot retry or complete after a session switch. Adds queue and fetch-engine regressions.


brandly:
@review

apps/inspect/src/log_data/fetchEngine.restart.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    version: 2,
+    status: "success",
+    eval: {
+      eval_id: name,
+      run_id: `run-${name}`,
+      task: "task",
+      model: "model",
+    },
+    sampleSummaries: [],
+  }) as unknown as LogDetails;
```
brandly: let's get rid of this type cast

apps/inspect/src/log_data/fetchEngine.restart.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    const getDetailsA = vi.fn(async (file: string) => {
+      await gate;
+      throw new Error(`failed: ${file}`);
+    });
+    const getDetailsB = vi.fn((file: string) =>
+      Promise.resolve(makeDetails(file))
+    );
+    const apiA = { get_log_details: getDetailsA } as unknown as ClientAPI;
+    const apiB = { get_log_details: getDetailsB } as unknown as ClientAPI;
```
brandly: get rid of these casts too?

apps/inspect/src/log_data/fetchEngine.restart.test.ts
brandly: Done — replaced with `testLogDetails`/`testEvalSpec` from the typed fixtures (merged main into the branch to pick them up).

apps/inspect/src/log_data/fetchEngine.restart.test.ts
brandly: Done — now `testClientAPI({ get_log_details })`, no casts.

apps/inspect/src/utils/workQueue.ts
```diff
@@ -206,6 +212,7 @@ export class WorkQueue<TInput, TOutput> {
   }
 
   clear() {
+    this.generation++;
```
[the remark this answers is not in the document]
brandly: Added the suggested doc comment on `clear()`.

PR #532, opened 2026-08-17
Force less ^4.8.1 to drop vulnerable image-size (dependabot alerts 73, 74)

i-am-marvin:
Clears both open dependabot alerts by adding a `less` entry to `pnpm.overrides`.

`image-size` has **no patched release** for either advisory (all published versions fall in the vulnerable range `<= 2.0.2`), but it only entered the tree via `less@4.5.1`. `less` >= 4.7.0 replaced `image-size` with `probe-image-size`, so forcing `less` to `^4.8.1` (published 2026-07-27, outside the 7-day soak window; 4.9.0 is still inside it) removes `image-size` from the lockfile entirely.

## Alerts fixed

- [dependabot alert 73](https://github.com/meridianlabs-ai/ts-mono/security/dependabot/73) — `image-size`, high — JXL and HEIF parsers allow denial of service through infinite loops (GHSA-5p2g-fcmc-qvqq)
- [dependabot alert 74](https://github.com/meridianlabs-ai/ts-mono/security/dependabot/74) — `image-size`, high — ICNS parser allows denial of service through an infinite loop (GHSA-w3rx-r6r6-pgpr)

## Deferred alerts

None.

## Verification

- `image-size@` no longer appears in `pnpm-lock.yaml` (only the unrelated `probe-image-size@7.3.0`)
- `pnpm audit`: no known vulnerabilities
- `pnpm check` (27 tasks), `pnpm build`, `pnpm test` (1139 tests, 106 files) all pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,8 @@
       "@babel/core": "^7.29.6",
       "undici": "^7.28.0",
       "dompurify": "^3.4.13",
-      "nanoid@3": "^3.3.18"
+      "nanoid@3": "^3.3.18",
+      "less": "^4.8.1"
     }
   },
   "manypkg": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -27,6 +27,7 @@ overrides:
   undici: ^7.28.0
   dompurify: ^3.4.13
   nanoid@3: ^3.3.18
+  less: ^4.8.1
 
 importers:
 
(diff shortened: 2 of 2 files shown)


Aprillion:
hmm, we are not using `less` in our code, it's a transient dependency of `typescript-plugin-css-modules` which is unused by CLI `tsc`, only referenced from https://github.com/meridianlabs-ai/ts-mono/blob/main/tooling/tsconfig/react.json for in-editor checks AFAICT

...does that dependency pay for its rent?

brandly:
I support removing `typescript-plugin-css-modules` and avoiding the issue that way! I don't think the plugin is wired up correctly, so we haven't really been using it anyway. cc @epatey since I know he was talking about the `image-size` vulnerability last week.

epatey:
Thanks. Agreed. Merged.

PR #544, opened 2026-08-18
Consolidate Scout score value rendering

brandly:
## Summary

- add Scout's fixed-preview `expandable={false}` behavior and array-compatible `className` support to the shared transcript `ScoreValue`
- migrate both Scout consumers to `@tsmono/inspect-components/transcript`
- delete Scout's duplicate `ScoreValue` implementation
- cover default expandable and fixed non-expandable object-score behavior with shared component tests

## Why

Scout and `@tsmono/inspect-components` carried near-identical `ScoreValue` implementations. Scout's copy had evolved two small API additions, so shared behavior changes could drift between the apps.

## Validation

- `pnpm -C packages/inspect-components test` — 918 passed, 2 skipped
- `pnpm -C packages/inspect-components lint`
- `pnpm -C packages/inspect-components typecheck`
- `pnpm -C apps/scout test` — 210 passed
- `pnpm -C apps/scout lint`
- `pnpm -C apps/scout typecheck`
- `pnpm -C apps/scout build:app`

## Manual QA

Tested a local Scout Vite page using the real shared `ScoreValue` and Scout `ScoreColumn` under Scout's theme and component providers:

- simple score rendered inline as `pass`
- expandable object score initially showed 2 of 5 rows and a `3 more...` control
- clicking `3 more...` showed all 5 rows and a `less...` control; clicking `less...` restored the 2-row preview
- Scout's non-expandable score column showed exactly 3 of 5 rows plus `All scores (5)`, with no internal more/less toggle
- no browser console errors occurred during the QA page checks

The temporary QA harness was removed before committing.


packages/inspect-components/src/transcript/ScoreValue.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+  ComponentNavigationProvider,
+} from "@tsmono/react/components";
+import {
+  ComponentStateHooks,
+  ComponentStateProvider,
+} from "@tsmono/react/state";
+
+import { ScoreValue } from "./ScoreValue";
+
+const icons: ComponentIcons = {
```
brandly: don't we have some `ComponentIcons` instance for testing that multiple tests utilize?

packages/inspect-components/src/transcript/ScoreValue.tsx
```diff
@@ -7,18 +7,37 @@ import { isRecord } from "@tsmono/util";
...
+  expandable = true
+): ReactNode => {
   if (Array.isArray(value)) {
     return value.join(", ");
   } else if (isRecord(value) && typeof value === "object") {
+    if (maxRows != null && !expandable) {
+      const visibleEntries = Object.fromEntries(
+        Object.entries(value).slice(0, maxRows)
+      );
+      return <MetaDataGrid entries={visibleEntries} />;
```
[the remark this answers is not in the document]
brandly: Done in 3056061e: the fixed-preview behavior is clarified in the prop docs and why-comment, with nested-group regression coverage.

packages/inspect-components/src/transcript/ScoreValue.test.tsx
brandly: Good call—1c403820 exports the existing `testIcons` through `@tsmono/react/testing` and reuses it here instead of copying the icon map.

PR #537, opened 2026-08-18
pnpm 11 and remove unused dependencies

Aprillion:
* globally installed pnpm 11 falled back to project's pnpm 10 only after showing a scary-sounding warning as if it didn't resolve security overrides => let's use project pnpm 11 to avoid the misleading warning:
   Screenshot 2026-08-17 130018
* removal of `typescript-plugin-css-modules` inspired by https://github.com/meridianlabs-ai/ts-mono/pull/532 because we don't use `less`, discussed in that PR and in https://inspectcommunity.slack.com/archives/C0B265TQQAF/p1786984951590469?thread_ts=1786982767.199109&cid=C0B265TQQAF
  * it was imperfectly wired so not even detecting *a lot* of CSS classname mismatches => candidate fix in follow-up PR
* clean up of package.json "unused" dependencies and regenerated lock file to drop the unused optional peer deps (ehm, `less`) and that includes in-range version updates that meet the minimum age of other deps - I had to test the app anyway so it sounded useful (..most "removed" stuff is still used by other packages and Fable was slightly overselling the benefits but it was still useful to clean up IMHO)
* I tested the web app with latest BE from inspect_ai, inspect_scout in browser, and the VS Code extension in Cursor on Win 11 and in local hawk (on top of https://github.com/METR/hawk/pull/1320)

<details>
<summary>
PR description from Fable
</summary>

**Title:** Pin pnpm 11 and remove unused dependencies (drops less/sass/stylus from the tree)

- **8e736f73** — pnpm 11 stops reading the `pnpm` field in package.json, so
  pinning 11.22.0 requires moving `overrides`/`peerDependencyRules` to
  pnpm-workspace.yaml first (otherwise only lockfile inertia enforces them and
  the next dep change silently drops them). Explicit `allowBuilds` (all false =
  unchanged behavior); dependabot-fix skill updated. Carries #532's
  `less ^4.8.1` override along (behavior-preserving; lockfile byte-identical).
- **8f49d870** — remove verified-unused deps. Headline: typescript-plugin-css-modules
  (editor-only, never correctly wired — maintainer concurrence:
  meridianlabs-ai/ts-mono#532 comment by brandly, 2026-08-17), whose hard deps
  were the only reason less/sass/stylus were in the tree. With less gone
  entirely, the `less ^4.8.1` override that merged #532 added for the
  image-size advisory (alerts 73/74) is dead — dropped here, along with the
  dead `allowBuilds: less` entry.
[...]

PR #553, opened 2026-08-20
CSS Modules Kit

Aprillion:
* type safety for CSS modules, follow-up from #537 removal of `typescript-plugin-css-modules` that transiently imported big unnecessary libs and was incorrectly wired so didn't actually detect the problems in most of the repo
* `css-modules-kit` chosen by "Fable research" but I tested that it actually works (normal tsc CLI (typescript language server in editors) for TSX files and eslint for CSS modules files .. no eslint for TSX because of false positives due to vscode-eslint engine not detecting live edits correctly .. CSS Modules Kit vsdode/cursor extension works fine for live edits and for `Ctrl+click` navigation (it even scrolls to correct place inside the CSS file)):
    Screenshot 2026-08-20 125020
* ~300 rules removed 😰 extensive before/after tests by Fable, review by multiple agents, and manual tests also in cursor extension on Win 11 and in hawk
  * full Fable's report with screenshots in https://peter.hozak.info/claude/css-modules-kit/ (note for agentic reviewers: make a list of independent concerns from code first, then look at the report if everything is addressed already or if you want to flag something extra)
  * 1 crash bug fix needed to test the touched CSS included in this PR (3rd commit):
      Screenshot 2026-08-20 135307
  * I reviewed the whole diff and agreed with the remaining conventions as explained by Fable (the comments and commit messages may have needed 1 or 2 rounds of deslopification but look very useful to me now 🤞)
  * hawk testing required also #554


.prettierignore
```diff
@@ -3,4 +3,5 @@ dist/
 lib/
 .turbo/
 pnpm-lock.yaml
-*.md
\ No newline at end of file
+*.md
+*.module.css.d.ts
```
brandly: this is ignored up here. do we actually need a bunch of individual files like `packages/zustand-devtools/.prettierignore`?

tooling/eslint-config/base.js
```diff
@@ -1,12 +1,38 @@
...
+// error-typed styles.<class> expressions the no-unsafe rule family
+// false-positives, duplicating the precise ts(2339) tsserver already shows
+// live. No editor-host detection is reliable across all eslint servers, so
+// the gate is inverted: the family is off unless the batch pipeline opts in.
+// Every package's `lint` script sets TSMONO_TYPED_LINT=1 inline, so
+// `pnpm lint`, per-package/--filter lint runs, turbo, and CI all enforce it;
+// editor servers (which load ESLint via its API, never through the scripts)
+// don't. Set the var in an editor server's env to force the rules on there.
+const typedLint =
+  !!process.env.TSMONO_TYPED_LINT && process.env.TSMONO_TYPED_LINT !== "0";
```
brandly: This is unfortunate! Seems like [the underlying issue](https://github.com/microsoft/vscode-eslint/issues/1774) won't be addressed any time soon.

brandly (approved):
Maybe it's worth seeing if we have extra `.prettierignore` files we don't need, but this looks great overall.

I really like [the QA report](https://peter.hozak.info/claude/css-modules-kit/) you included. What are you using to generate and publish those reports? I'd love to have that for every UI update.

Thanks for doing this. It's always great to have more type safety.

brandly:
@Aprillion do you mind doing some manual testing in the vscode viewer? These styles are all impacted by your vscode theme, so it's worth theme-switching and making sure nothing looks strange.

Aprillion:
I did theme switching in Cursor already... but it was a "few" fixes ago, and not by my laptop tonight (Prague timezone), will retest tomorrow morning.

Aprillion:
ah, package level similar eslint configs were necessary, but I haven't asked/tested prettierignore duplication, will check

Aprillion:
the report is by Fable with naked prompting, started as md file, then md file not in /tmp folder, then converted to html for working screenshot comparison, then some swearing about duplicated out of date reports, to deslop it and to not write a novella abot its heroic journey but just the final state description ... not extracted into a skill yet

.prettierignore
Aprillion: yeah, looks like these are needed at the moment:
> removing a package's `.prettierignore` does make `prettier --check .` flag its generated d.ts — so they're not redundant; each package runs `prettier --check .` from its own dir, and Prettier (unlike ESLint/git) doesn't cascade to ancestor `.prettierignore` files

..I imagine we could replace the per-package prettier with global one, should I try that in a follow-up PR?

Aprillion:
theme switching looks OK to me, including system theme in browser and event colors toggle and nothing sus in cursor extension:

https://github.com/user-attachments/assets/523027c5-ba90-4743-bfce-63614b6aa1b2

.prettierignore
brandly: Okay! Maybe it could lead to some config simplification, but I'm not too worried about these extra `.prettierignore`

brandly:
@Aprillion thank you!! Great work

PR #554, opened 2026-08-20
Fix library builds emitting runtime require("react") that black-screens browser consumers

Aprillion:
* by Fable, but used for testing hawk in #553

TanStack Table v9 pulls in @tanstack/react-store, which imports the CJS-only use-sync-external-store package. Library builds externalize react, so Rolldown kept that package's internal require("react") as a runtime __require that throws in browsers ("Calling 'require' for 'react' in an environment that doesn't expose the 'require' function"), breaking embedders that consume lib/index.js (e.g. hawk).

Externalize use-sync-external-store in the library builds and declare it as a direct dependency of both apps, so the emitted library is pure ESM and the consumer's bundler performs the CJS interop (the same way zustand/react-redux consumers get it).

### DIFF
--- a/apps/inspect/package.json
+++ b/apps/inspect/package.json
@@ -117,6 +117,7 @@
     "react": "^19.2.8",
     "react-dom": "^19.2.8",
     "react-router": "^8.3.0",
+    "use-sync-external-store": "^1.6.0",
     "zustand": "^5.0.14"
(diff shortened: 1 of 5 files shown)


brandly:
@review

brandly (approved):

PR #578, opened 2026-08-25
Rewrite compiler-unsupported syntax so 15 more files compile (React Compiler)

brandly:
Companion to #576. A repo-wide run of the reference `babel-plugin-react-compiler` (1.0.0) found 63 components/hooks bailing out of React Compiler, 19 of them on **syntax the compiler can't lower yet** (its `Todo` class) rather than rules-of-react violations. This PR rewrites those sites behavior-preservingly so the surrounding components compile.

**Result: bail-out sites 63 → 48, compiled functions 756 → 771.** Fifteen files now compile that previously ran completely unmemoized.

### The rewrites

| Blocker | Fix |
|---|---|
| `try/finally` | Where the catch never rethrows: cleanup moved after the block (equivalent). Where cleanup must survive a throw (`useColumnSizing`'s ref reset around `computeSizes`): rethrowing catch + trailing reset. |
| Value blocks (`\|\|`, `??`, `?.`, ternaries) and loops inside `try`/`catch` | Hoisted above the try, or extracted to module-level helpers. Only the genuinely throwing calls stay inside — e.g. `buildMetadataSet` keeps `serializeEntry` under the catch that classifies `MetadataParseError`. |
| Captured `idx++` / `++frames` in lambdas | Indexing arithmetic (`ScoreGrid`) or `+= 1` — the compiler lowers reassignment, just not `UpdateExpression`. |
| `??=`, computed `CallExpression` keys, `for await` in a hook, | Explicit `if`, hoisted const, module-level `collectExportText`. |

### Deliberately untouched

- **`useVisitId`** — its render-time module counter is rejected by the compiler regardless of syntax (`Globals`); the pattern is intentional, left as-is.
- **`VirtualList`, `TranscriptViewNodes`, `useChromeNavOwnership`** — these have deeper `PreserveManualMemo`/`Refs`/`Immutability` bail-outs that are judgment work, not mechanical; only their trivial syntax blockers are cleared here (they still bail, so their existing manual memoization remains load-bearing).
- The `Refs`/`Hooks`/`Suppression` bail-out classes (real rules-of-react flags) — separate follow-up material.

### Verification

- `turbo lint typecheck format:check test` green for both apps and all touched packages (inspect: 1185 tests, scout + packages all passing)
- Both apps build through the production SWC `reactCompiler` path
- Re-ran the Babel-reference bail-out report before/after; every touched file except the three "deliberately untouched" ones now compiles cleanly

Caveat: measurements use the Babel reference implementation; the production SWC Rust port's `Todo` list may differ at the margins, but these rewrites are neutral-to-positive under either.

No UI-visible changes (control-flow-equivalent rewrites only), so no screenshots.

https://claude.ai/code/session_01VdJMGhgZqwoxz6pxcVmMbN

---
_Generated by [Claude Code](https://claude.ai/code/session_01VdJMGhgZqwoxz6pxcVmMbN)_

### DIFF
--- a/apps/inspect/src/app/log-list/ViewerOptionsPopover.tsx
+++ b/apps/inspect/src/app/log-list/ViewerOptionsPopover.tsx
@@ -40,8 +40,6 @@ export const ViewerOptionsPopover: FC<ViewerOptionsPopoverProps> = ({
       setClearMessage("Failed to clear database");
       setTimeout(() => setClearMessage(null), 3000);
     }
-    // Not a finally clause: React Compiler can't lower try/finally, and the
-    // catch never rethrows, so running after the block is equivalent.
     setIsClearing(false);
   };
 
--- a/apps/inspect/src/app/log-list/ViewerOptionsPopover.tsx
+++ b/apps/inspect/src/app/log-list/ViewerOptionsPopover.tsx
@@ -39,9 +39,10 @@ export const ViewerOptionsPopover: FC<ViewerOptionsPopoverProps> = ({
       console.error("Failed to clear database:", error);
       setClearMessage("Failed to clear database");
       setTimeout(() => setClearMessage(null), 3000);
-    } finally {
-      setIsClearing(false);
     }
+    // Not a finally clause: React Compiler can't lower try/finally, and the
+    // catch never rethrows, so running after the block is equivalent.
+    setIsClearing(false);
   };
 
   return (
--- a/apps/inspect/src/app/log-view/title-view/EditMetadataDialog.tsx
+++ b/apps/inspect/src/app/log-view/title-view/EditMetadataDialog.tsx
@@ -362,8 +362,6 @@ export const EditMetadataDialog: FC<EditMetadataDialogProps> = ({
         setError(formatEditError(err));
       }
     }
-    // Not a finally clause: React Compiler can't lower try/finally, and the
-    // catch never rethrows, so running after the block is equivalent.
     window.clearTimeout(indicatorTimer);
     setSubmitting(false);
     inFlightRef.current = false;
--- a/apps/inspect/src/app/log-view/title-view/EditMetadataDialog.tsx
+++ b/apps/inspect/src/app/log-view/title-view/EditMetadataDialog.tsx
@@ -119,6 +119,17 @@ export const serializeEntry = (entry: MetaEntry): unknown => {
   }
 };
 
+// Module-level so the dialog stays compilable: React Compiler can't lower
+// loops inside try/catch, and serializeEntry must run under the catch that
+// classifies MetadataParseError.
+const buildMetadataSet = (entries: MetaEntry[]): Record<string, unknown> => {
+  const metadata_set: Record<string, unknown> = {};
+  for (const entry of entries) {
+    metadata_set[entry.key] = serializeEntry(entry);
+  }
+  return metadata_set;
+};
+
 const seedFor = (type: NewType): unknown => {
   switch (type) {
     case "string":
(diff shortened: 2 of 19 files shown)


apps/inspect/src/app/log-list/ViewerOptionsPopover.tsx
```diff
@@ -39,9 +39,10 @@ export const ViewerOptionsPopover: FC<ViewerOptionsPopoverProps> = ({
       console.error("Failed to clear database:", error);
       setClearMessage("Failed to clear database");
       setTimeout(() => setClearMessage(null), 3000);
-    } finally {
-      setIsClearing(false);
     }
+    // Not a finally clause: React Compiler can't lower try/finally, and the
+    // catch never rethrows, so running after the block is equivalent.
```
brandly: Remove this whole comment

apps/inspect/src/app/log-view/title-view/EditMetadataDialog.tsx
```diff
@@ -312,29 +323,32 @@ export const EditMetadataDialog: FC<EditMetadataDialogProps> = ({
...
-          reason: reason.trim() || undefined,
-          metadata: {},
-          timestamp: new Date().toISOString(),
-        },
+        provenance,
       });
       setShowing(false);
-      onSaved?.();
+      if (onSaved) {
+        onSaved();
```
brandly: This really affects the compiler? I'm surprised

apps/inspect/src/app/log-view/title-view/EditMetadataDialog.tsx
```diff
@@ -347,11 +361,12 @@ export const EditMetadataDialog: FC<EditMetadataDialogProps> = ({
       } else {
         setError(formatEditError(err));
       }
-    } finally {
-      window.clearTimeout(indicatorTimer);
-      setSubmitting(false);
-      inFlightRef.current = false;
     }
+    // Not a finally clause: React Compiler can't lower try/finally, and the
```
brandly: Remove this comment too

brandly:
Addressed the review nit in f096291: `mergeCalculatedSizing` moved to the shared `app/components/columnSizing` module (alongside `clampSize` in `types.ts`, exported from the barrel), and both `useColumnSizing` hooks now import it instead of carrying their own copy. Verified scout lint/typecheck/format/test green and both hooks still compile under React Compiler.

(Replying at PR level — a pending review on this PR blocks threaded replies from me.)

---

apps/inspect/src/components/DownloadLogButton.tsx
```diff
@@ -33,11 +33,12 @@ export const DownloadLogButton = ({
     } catch (error) {
       console.error("Failed to download log:", error);
       setDownloadState("error");
-    } finally {
-      setTimeout(() => {
-        setDownloadState("idle");
-      }, 1250);
     }
+    // Not a finally clause: React Compiler can't lower try/finally, and the
+    // catch never rethrows, so running after the block is equivalent.
```
brandly: Remove comment

apps/scout/src/app/validation/components/CopyMoveCasesModal.tsx
```diff
@@ -275,11 +278,12 @@ export const CopyMoveCasesModal: FC<CopyMoveCasesModalProps> = ({
...
     } catch (err) {
-      const message = err instanceof Error ? err.message : "Operation failed";
-      setError(message);
-    } finally {
-      setIsProcessing(false);
+      setError(errorMessage(err, "Operation failed"));
     }
+    // Not a finally clause: React Compiler can't lower try/finally. Every
+    // early return inside the try already resets isProcessing itself, and
+    // the catch never rethrows, so running after the block is equivalent.
```
brandly: Remove comment

packages/inspect-components/src/transcript/TranscriptViewNodes.tsx
```diff
@@ -444,7 +444,7 @@ export const TranscriptViewNodes = forwardRef<
     container.addEventListener("keydown", abort, { passive: true });
     const settle = () => {
       if (cancelled) return;
-      attempts++;
+      attempts += 1;
```
brandly: This affects the compiler?

brandly:
Addressed the review comments in 69c0e9a (replying at PR level — a pending draft review on the repo blocks threaded replies from me):

**Comment cleanup** — removed the "Not a finally clause" explainer comments at all four flagged sites, plus the identical one in `EditTagsDialog.tsx` for consistency. Five sites total, comments only, no code change.

**`EditMetadataDialog.tsx` — "`if (onSaved) onSaved()` really affects the compiler?"** — Yes, but only because it sits *inside the `try` block*. `onSaved?.()` is fine anywhere else; inside a `try`/`catch` the compiler can't lower "value blocks" (`?.`, `??`, `||`, ternaries — expressions that compile to branching control flow), and one unsupported expression bails the entire component. Verified empirically: with `?.` restored there, the reference compiler reports `Todo: Support value blocks (conditional, logical, optional chaining, etc) within a try/catch statement` and the whole dialog compiles to zero memo slots.

**`TranscriptViewNodes.tsx` — "`attempts += 1` affects the compiler?"** — `attempts++` does: `++`/`--` on a variable captured by a closure is unsupported (`Todo: Handle UpdateExpression to variables captured within lambdas`), while `+= 1` lowers fine. Honest caveat: this particular component *still* bails today on a separate refs-during-render issue, so this line alone doesn't unlock it — it just clears the syntax-level blocker so the remaining diagnostic points at the real problem. Happy to revert this one line if you'd rather keep the file untouched until the refs issue is dealt with.

---

PR #592, opened 2026-08-28
Remove eleven stale pnpm overrides and document the rest

brandly:
Started as "is this babel config even used?" and turned into an audit of everything we override.

`apps/inspect/babel.config.js` was dead. It referenced `@babel/preset-env`, `@babel/preset-typescript` and `@babel/plugin-transform-modules-commonjs`, none of which are installed, so loading it would have failed on a missing preset. Nothing loaded it. We build with `@vitejs/plugin-react-swc` and test with vitest. It was copied in with the original inspect_ai import and never touched since.

That led to the overrides. There were 21, now there are 10.

Eleven came out, in five commits:

- `lodash`, `@microsoft/api-extractor`, `fast-uri`, `immutable` no longer resolve at all. Their only trace in the lockfile was the overrides echo.
- `yaml` and `@codemirror/state` were floors every consumer's own range had already cleared. `@codemirror/state` was also rewriting apps/inspect's declared `^6.7.1` down to `^6.5.4`.
- `seroval` and `seroval-plugins` were pushing past the `~1.5.4` cap solid-js declares. The advisory behind them is fixed in 1.4.1, and 1.5.6 clears it, so dropping them keeps us patched and inside solid-js's range.
- `ws` and `undici` came out after review flagged them as the only two without a why comment. Working out what to write showed neither should survive. `undici` had inverted: jsdom declares `^8.9.0` and the `^7.28.0` pin capped it a major below that. `ws` was inert and misleading, its `^8.20.1` floor now sits below GHSA-96hv-2xvq-fx4p (patched in 8.21.0), while storybook's own `^8.21.1` is what actually keeps us on a safe 8.21.3.
- `glob ^10.5.0` had inverted. It was added upstream to drag a vulnerable glob 7 up. The one consumer left, storybook's react-docgen plugin, now asks for `^13.0.1`, so the pin was holding us three majors down on a release whose own npm deprecation notice says old glob versions "contain widely publicized security vulnerabilities". Removing it also drops 13 transitive packages that only glob 10 needed.

The last commit documents the survivors. The rationale lived in commit messages spread across inspect_ai, inspect_scout and here, which is no good to anyone reading the file. Now that overrides live in yaml instead of json we can just write it down.

I verified each removal by diffing the full resolved package set before and after, not by reading ranges. That caught me being wrong about seroval: I had filed it as inert, and removing it downgraded 1.6.2 to 1.5.6. A `>=` override is not a floor. It replaces the consumer's range and resolves to the newest published version.

## Current behavior

Seven overrides do nothing. Two hold a package below what its consumer asks for. Two push a package past a cap its consumer declares. Every install prints a deprecation warning for `glob@10.5.0`.

## New behavior

Ten overrides, each verified as load-bearing and commented with why it exists and what would let someone delete it. No deprecation warning.

Only `glob`, `seroval` and `undici` change a resolved version. Everything else is metadata: 888 packages to 875, all from glob 10's dependency tail.

## Breaking changes

None. `glob` is not declared by any workspace package and reaches us through `@storybook/react-vite` devDependencies. `seroval` moves to what solid-js asks for, and `undici` to what jsdom asks for.

## Other information

`pnpm audit` reports no known vulnerabilities. `pnpm check` 33/33, 1234 tests pass, suppressions ledger unchanged.

One thing worth flagging separately: the repo root is not format-checked. Each workspace runs `prettier --check .` in its own directory, so `pnpm-workspace.yaml`, `turbo.json` and the root `package.json` are covered by nothing. `pnpm check` passed on my yaml edit without ever looking at it. I ran prettier on the file directly instead. Not fixed here.

### Agent review

- Reviewer: Claude Opus 5, same context as the author, no separate review pass
- Verification was empirical rather than a review pass. Every removal was checked by diffing the resolved package set from `pnpm-lock.yaml` before and after `pnpm install`, plus `pnpm audit`, `pnpm check` and the full test suite. That process caught two of my own wrong calls: seroval, described above, and a comment crediting `minimatch ^3.x` to two consumers when there are three (`eslint-plugin-jsx-a11y` is also on that line).
- The automated `review` job did a fresh-context pass and verified the claims against the lockfile independently. Its one nit, that `ws` and `undici` lacked a why comment, turned out to point at two more stale overrides rather than two missing comments.


brandly:
Good catch, and it went further than a missing comment. I went to write the two `# why` lines and found neither override should still be there.

`undici` had inverted, same as `glob`. Its only consumer is jsdom 30.0.1, which declares `undici: ^8.9.0`, so the `^7.28.0` pin was capping it a major below what jsdom asks for. Removing it resolves undici 8.10.0 and clears every advisory the pin carried. The newest, GHSA-4cwx-7wf7-3272 (high), is fixed in 7.29.0 and 8.9.0.

`ws` was inert, and misleading about it. The `^8.20.1` floor encodes GHSA-58qx-3vcg-4xpx, patched in 8.20.1. But storybook 10.5.8 declares `ws: ^8.21.1` and that is what actually pins us to 8.21.3. Meanwhile the override floor now sits below GHSA-96hv-2xvq-fx4p (high, memory exhaustion DoS), patched in 8.21.0. So the pin was not protecting anything, and if it ever became the binding constraint it would permit a vulnerable 8.20.x.

Both removed in e9a30be7. Resolved-set delta is undici 7.29.0 to 8.10.0 and nothing else. ws stays at 8.21.3. `pnpm audit` clean, check 33/33, tests pass.

That is 21 overrides down to 10. Description updated.

Issue #555, opened 2026-08-20
Data isn't validated at the system boundary — parsed types lie, forcing defensive guards


epatey:
## Problem

Data entering the client is cast to generated/declared types without runtime validation:

- eval log files and sample data (`ModelEvent`, `EvalSpec`, `EvalResults`, `LogHeader`, ...)
- journal files (`_journal/start.json` is `JSON.parse` + `as LogStart`; timelineData notes "journal entries are cast, not validated")
- server/API responses (generated `api-types`)
- persisted webview/localStorage state (VS Code restored sample handles, zustand-persisted slices)

Old or partial files omit fields the types declare required, so the types overstate what's actually there. Downstream code compensates with defensive `?.`/guards that the compiler and `@typescript-eslint/no-unnecessary-condition` correctly flag as impossible — per the types.

## Cost today

PR #404 enabled `no-unnecessary-condition`. An audit of the 179 "unnecessary" optional chains found **72 load-bearing defensive guards** that exist only because the boundary doesn't validate. Each now carries a suppression comment:

`-- intentional: data isn't validated at the wire; old files may omit type-required fields`

Every future rule violation on parsed data will need the same treatment, and any agent or human who trusts the types can reintroduce crashes (removing these guards broke 7 tests: `event.config`, `event.output`, `provenance.metadata`, ...).

## Goal

Validate or normalize at the boundary (parse, don't cast) so that types downstream are honest:

- parse-time validation/normalization for eval log + journal data (fill defaults or mark fields optional to match reality)
- same for persisted state and API responses where the schema can drift

Done means: the `intentional: data isn't validated at the wire` suppressions can be deleted (grep for that string), `no-unnecessary-condition` findings on parsed data become trustworthy, and defensive `?.` is no longer needed outside the boundary layer.

PR #593, opened 2026-08-28
chore: suppression burn-down (floating promises + file-wide condition disables)

epatey:
Suppression-ledger burn-down: the two biggest `no-floating-promises` hotspots, plus conversion of the 7 file-wide `no-unnecessary-condition` disables to per-line. Ledger: 511 → 513 total but **reason-less 338 → 298**, and zero file-wide disables of this rule remain (each previously counted as 1 while hiding 45 real sites and suppressing all future ones).

## fetchEngine.test.ts (−24, file now suppression-free)

- `releaseAll()` in the test fake is now synchronous (setInterval drain, same 100×5ms schedule), so bare fire-and-forget calls are legal — 13 suppressions gone.
- The 11 backgrounded `ensure(previewed)` calls are now awaited. `ensure` with `depth: "previewed"` resolves at enqueue (see the `ensure` docstring), so this is a no-op at runtime — 11 suppressions gone.
- The one site that awaited the full 500ms drain (F5) now waits for the in-flight mock call to settle (`vi.waitFor` + `toHaveResolvedTimes`) plus two ticks, instead of sleeping.

## useValidations.ts (−6, 3 kept with reasons)

Split by mutation kind (query-core's `invalidateQueries`/`cancelQueries` never reject — refetch/cancel errors are caught internally unless `throwOnError` — so this is purely about whether `isPending` covers the refetch):

- **Non-optimistic** (delete case, bulk delete, delete set, rename set): `await queryClient.invalidateQueries(...)` in async `onSuccess`. Mutation stays pending until the refetch lands, so a deleted row / old name can't linger after the spinner stops — matches how ValidationPanel gates its buttons on `isPending`. Bulk delete's invalidation moves from `mutationFn` to `onSuccess` (its `succeeded > 0` guard was redundant: all-failed throws).
- **Optimistic** (create set, update case): invalidation stays fire-and-forget — background reconciliation is the design — now carrying `-- reason` directives. `useUpdateValidationCase.onMutate` now awaits its two `cancelQueries` (the documented optimistic-update pattern; create-set already did).

## File-wide `no-unnecessary-condition` disables → per-line (7 headers hiding 45 sites)

`eventText.ts`, `timeline/core.ts`, `ModelEventView.tsx`, `fixtureTimeline.test.ts`, scout `TranscriptBody.tsx` + `ScannerSidebar.tsx`, inspect `log-list/grid/columns/hooks.tsx`. Conservative pass — **guards on wire/API data were kept, not deleted**:

- **39 per-line directives with reasons** naming why each guard stays: unnormalized API surfaces (scout responses, scan status), types-lie sites pending #555 normalizer verification (e.g. `ModelEventView.tsx`'s documented empirically-undefined `event.output`), and `getValue<T>()`'s unchecked generic in the grid.
- **5 sites deleted** — the only ones provable by local reasoning: a truthiness check on a `.filter()` result (always an array) in `timeline/core.ts`, and in `fixtureTimeline.test.ts` a dead undefined-guard plus three exhaustive-union comparisons (`x === "span" || x === "tool"` over a `"span" | "tool"` type).

Follow-up (not this PR): per-site #555 verification to actually remove the kept guards — check/add normalizer fills, then delete downstream guards.

## Verified

- inspect: full `vitest run` 1234 tests pass (fetchEngine 52/52, faster — no 500ms sleep)
- scout: full `vitest run` 270 tests pass
- inspect-components: full `vitest run` 953 tests pass
- eslint clean on all touched files
- `node scripts/check-suppressions.mjs` passes

### DIFF
--- a/apps/inspect/src/app/log-list/grid/columns/hooks.tsx
+++ b/apps/inspect/src/app/log-list/grid/columns/hooks.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-unnecessary-condition */
 import clsx from "clsx";
 import { useCallback, useMemo } from "react";
 
@@ -385,6 +384,7 @@ export const useLogListColumns = (
         accessorFn: (row) => row.totalSamples,
         cell: ({ getValue }) => {
           const value = getValue<number | undefined>();
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- getValue's type argument is an unchecked assertion; row data may contain nulls the type omits
           if (value === undefined || value === null) {
             return <EmptyCell />;
           }
@@ -401,6 +401,7 @@ export const useLogListColumns = (
         accessorFn: (row) => row.completedSamples,
         cell: ({ getValue }) => {
           const value = getValue<number | undefined>();
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- getValue's type argument is an unchecked assertion; row data may contain nulls the type omits
           if (value === undefined || value === null) {
             return <EmptyCell />;
           }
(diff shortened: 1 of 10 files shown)


epatey:
@review

PR #598, opened 2026-08-31
Split suppression delta comment into a pull_request_target workflow so it posts on fork PRs

epatey:
## Why

The "Comment ledger delta" step ran inside ci.yaml's `suppressions` job on `pull_request`. On fork PRs that event's `GITHUB_TOKEN` is read-only, so the comment silently failed to post: a properly-ledgered new suppression from a fork landed green with nothing drawing a reviewer's eye to the ledger diff. inspect_ai's port of this gate (UKGovernmentBEIS/inspect_ai#5136) hit and fixed the same problem; this brings the fix back, plus the other language-agnostic improvements that came out of that PR's review rounds.

## What

- ci.yaml `suppressions` job: comment step and its `pull-requests: write` removed. Enforcement (gate check + script tests) stays on `pull_request` with `contents: read`.
- New `.github/workflows/suppressions-comment.yml` on `pull_request_target` posts the sticky delta comment, so it works on fork PRs too. Same `<!-- suppressions-delta -->` marker logic as before.

## Security

`pull_request_target` carries a write token even on fork PRs, which is a remote-code-execution footgun if the job ever runs PR-controlled code. The workflow keeps three invariants, documented in a header block in the YAML:

1. Only base code runs. `actions/checkout` has no `ref`, so it checks out the base branch: the delta script and base ledger are the trusted copies. A PR's edits to the script take effect only after merge.
2. PR content is data, never code. The PR's `suppressions.json` is read with `git fetch` + `git show FETCH_HEAD:suppressions.json`, which reads a blob and executes nothing. The PR head is never checked out.
3. No dependency installs. The delta script is node-builtins-only (it imports only `node:fs`, `node:url`, and its sibling gate script); `pnpm install` would execute PR-controllable lifecycle and dependency scripts.

The delta comment also now escapes ledger text (file paths, rule names) before rendering it into the comment table, since that text can come from a fork.

## Ported from the inspect_ai version

Small, language-agnostic improvements its review rounds produced:

- Delta comment escapes untrusted cell text (HTML, `|`, newlines).
- Warning heading and footer when the reason-less count grows even though the total does not (previously rendered as a plain non-warning change).
- Delta script reuses `diffLedgers`/`totals` from the gate script instead of duplicating them, and exports a testable `render()`.
- `diffLedgers` output is sorted, so check-mode error messages and comment rows come out in stable order.
- Unknown CLI args are rejected (a typo like `--updat` used to silently run check mode instead of update).
- Bootstrap `--update` says the ratchet was not applied, so deleting the ledger is not an invisible bypass.
- `--update` prints why it refuses on reason-less growth instead of exiting silently after the UNDESCRIBED lines.
- `git ls-files --cached --others --exclude-standard` with an existence filter: an update run before `git add` sees newly authored files, and unstaged deletions do not crash the scan.
- Tests: 6 delta render tests (new file) and 3 CLI tests against a throwaway git repo (bootstrap, ratchet refusal, unknown args). CI runs both test files.

Not ported, flagged for completeness:

- `--allow-growth` update override: inspect_ai needs it for merges from its upstream that bring in reason-less suppressions on lines it must not edit. ts-mono has no such merge flow. Easy to add later if one appears.
- `release/**` branch triggers on both workflows: ts-mono has no release branches.
- The mypy/ruff directive-semantics fixes (tokenize handling, legacy type comments, file-wide spelling rules) are Python-specific, N/A here.

## Testability caveat

A `pull_request_target` workflow runs from the base branch's copy, so the new workflow does not run for this PR. It activates for PRs opened after this merges. The delta body logic was verified locally instead: add, remove, unchanged, and missing-file cases all render correctly via the CLI, plus the unit tests above.

## Verification

- `pnpm check` passes (manypkg, suppressions gate, lint, typecheck, format).
- `pnpm test` passes.
- `node --test scripts/check-suppressions.test.mjs scripts/suppressions-pr-delta.test.mjs`: 28 pass.
- Gate check mode on the real repo: ledger matches (727 suppressions in 308 files), confirming the `ls-files` change does not alter a clean checkout's scan.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,11 +39,14 @@ jobs:
       - name: Check linting
         run: pnpm lint
 
+  # Enforcement only: it runs fine with the read-only token fork PRs get.
+  # The PR delta comment is posted by suppressions-comment.yml
+  # (pull_request_target), which needs a write token — keep the two
+  # concerns in their separate workflows.
   suppressions:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      pull-requests: write
     steps:
       - uses: actions/checkout@v7
 
@@ -56,34 +59,8 @@ jobs:
       - name: Check suppressions ledger
         run: node scripts/check-suppressions.mjs
 
-      - name: Test the gate script
-        run: node --test scripts/check-suppressions.test.mjs
-
-      - name: Comment ledger delta on PR
-        if: github.event_name == 'pull_request'
-        continue-on-error: true
-        env:
-          GH_TOKEN: ${{ github.token }}
-          GH_REPO: ${{ github.repository }}
-          BASE_REF: ${{ github.base_ref }}
-          PR: ${{ github.event.pull_request.number }}
-        run: |
-          git fetch --depth=1 origin "$BASE_REF"
-          git show "origin/$BASE_REF:suppressions.json" > /tmp/base-suppressions.json 2>/dev/null \
-            || echo '{}' > /tmp/base-suppressions.json
-          body=$(node scripts/suppressions-pr-delta.mjs /tmp/base-suppressions.json suppressions.json)
-          # Sticky comment: match on the marker, not --edit-last, so other
-          # bot comments on the PR are never clobbered.
-          cid=$(gh api "repos/$GH_REPO/issues/$PR/comments" --paginate \
-            -q '.[] | select(.body | startswith("<!-- suppressions-delta -->")) | .id' | head -1)
-          if [ -n "$body" ] && [ -n "$cid" ]; then
-            gh api -X PATCH "repos/$GH_REPO/issues/comments/$cid" -f body="$body" > /dev/null
-          elif [ -n "$body" ]; then
-            gh pr comment "$PR" --body "$body"
-          elif [ -n "$cid" ]; then
-            gh api -X PATCH "repos/$GH_REPO/issues/comments/$cid" \
-              -f body="$(printf '<!-- suppressions-delta -->\nThis PR no longer changes the suppression ledger.')" > /dev/null
-          fi
+      - name: Test the gate scripts
+        run: node --test scripts/check-suppressions.test.mjs scripts/suppressions-pr-delta.test.mjs
 
   typecheck:
     runs-on: ubuntu-latest
--- a/.github/workflows/suppressions-comment.yml
+++ b/.github/workflows/suppressions-comment.yml
@@ -1,0 +1,89 @@
+# Posts the sticky suppression-ledger delta comment on PRs, fork PRs
+# included — see scripts/suppressions-pr-delta.mjs and the suppressions
+# section in CONTRIBUTING.md.
+#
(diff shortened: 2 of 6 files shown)


epatey:
@auto

i-am-marvin:
**Claude finished @epatey's task in 8m 30s** —— [View job](https://github.com/meridianlabs-ai/ts-mono/actions/runs/33412815243)

---
Addressed both non-blocking review nits on `scripts/suppressions-pr-delta.mjs` and pushed cfcd701f.

- [x] Merge `origin/main` into this branch (already up to date — no merge commit needed)
- [x] Escape `[` and backticks in `cell()` so fork-controlled ledger text can't render live links/images
- [x] Make `main()` require exactly two path args and fail loudly otherwise
- [x] Extend tests for both changes
- [x] Run tests + `pnpm check` + `pnpm test`
- [x] Commit, push, resolve the two review threads

### What changed

**Cell escaping** (`scripts/suppressions-pr-delta.mjs:27`): `cell()` now also replaces `[` → `[` and `` ` `` → ```. Without an opening bracket, a `](url)` tail is inert text, so fork-controlled ledger text like `` can no longer render a live image/link inside the bot comment; the backtick escape keeps stray code spans from forming.

**Strict CLI args** (`scripts/suppressions-pr-delta.mjs:87`): `main()` now requires exactly two path args and exits 2 with a usage message otherwise, matching the gate script's strict arg handling — a future workflow edit that mis-invokes the script fails loudly instead of printing nothing and silently resetting the sticky comment to "no longer changes the ledger". Missing *files* stay tolerated (`load()`'s `{}` fallback), which the workflow relies on for branches that predate the ledger.

**Tests**: updated the escape test for the new entities and added three tests — link/image syntax cannot form in a cell, the CLI rejects 0/1/3 args with exit 2, and missing ledger files still render as "no change" with exit 0. Suite is now 31 tests, all passing.

### Verification

- `node --test scripts/check-suppressions.test.mjs scripts/suppressions-pr-delta.test.mjs`: 31 pass, 0 fail
- `pnpm check`: 33/33 tasks pass (manypkg, suppressions gate, lint, typecheck, format)
- `pnpm test`: 9/9 tasks pass

Both review threads are resolved. Branch was already up to date with `origin/main`, so no merge commit was needed.
 · [Branch](https://github.com/meridianlabs-ai/ts-mono/tree/supp-comment-split)

i-am-marvin:
@review

i-am-marvin:
@review

i-am-marvin:
⚠️ The auto review-fix run did not complete:

> API Error: 400 You have reached your specified workspace API usage limits. You will regain access on 2026-09-01 at 00:00 UTC.

If this looks transient (model API capacity/overload), re-trigger with `@review`; otherwise it likely needs a human.

epatey:
@review

i-am-marvin:
🤖 auto: my last fix round pushed no new changes, so I'm not making progress on the reviewer's remaining feedback. Removing the `auto` label and handing this to a human. Re-add it to let me keep iterating.

PR #599, opened 2026-08-31
Show cost in the log viewer usage panels

dragonstyle:
## What this does

When an eval log records model costs (inspect_ai computes these whenever pricing is configured via `set_model_cost()` / `--model-cost-config`), the viewer now shows dollars alongside tokens:

- **Eval Models tab and sample Usage tab**: a "Cost" item in the panel header with the total, per-model cost beneath each token total, and an "avg cost / sample" figure in the per-sample column.
- **Transcript model events**: per-call cost appended to the input/output summary line.
- **Partial pricing**: if some models used tokens but recorded no cost (mock/local/unpriced), the total renders as a lower bound (`≥ $4.56`) with an explanatory tooltip, so it never silently under-reports.
- **Old logs and unpriced runs render exactly as before** — no cost UI appears when no cost was recorded.

Formatting adapts to magnitude: `$0.0042` for sub-cent per-call values (fixed two decimals would render `$0.00`), `$1,234.56` above a dollar, always two significant digits below a dollar so `$0.10` never truncates to `$0.1`.

## Implementation notes

- `formatCurrency` added to `@tsmono/util` (11 table-driven test cases).
- `costSummary` helper in `inspect-components/usage` sums recorded costs and detects partial pricing (6 test cases); render coverage for the table in `ModelTokenTable.test.tsx`.
- `ModelUsageData` gained `total_cost` — the field already existed in the generated log schema (`ModelUsage.total_cost`); the hand-rolled interface had drifted. No schema changes, so the paired inspect_ai PR is a submodule bump + dist rebuild only.

## Screenshots

Verified against a real run (4-sample agentic eval with `--model-cost-config`) and a synthetic log exercising the partial-pricing path. A light-mode screenshot of the sample Usage tab (real run, real costs) is attached below. Note the "before" state is directly visible in any log without recorded costs — the cost UI is fully conditional, so no-cost logs render identically to the current UI.

### DIFF
--- a/packages/inspect-components/src/usage/ModelTokenTable.module.css
+++ b/packages/inspect-components/src/usage/ModelTokenTable.module.css
@@ -74,6 +74,14 @@
   margin-top: 0.4rem;
 }
 
+.modelCost {
+  display: block;
+  font-size: 0.8125rem;
+  font-weight: 600;
+  color: var(--bs-secondary);
+  margin-top: 0.15rem;
+}
+
 .configSection {
   margin-top: 12px;
   padding-top: 10px;
@@ -201,6 +209,12 @@
   letter-spacing: 0.04em;
   text-transform: uppercase;
   margin-top: 0.15rem;
+}
+
+.perSampleCost {
+  display: block;
+  margin-top: 0.5rem;
+  color: var(--bs-body-color);
 }
 
 .swatchSmall {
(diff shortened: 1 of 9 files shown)


brandly:
> A light-mode screenshot of the sample Usage tab (real run, real costs) is attached below.

but there's no screenshot! i'd be interested in seeing the transcript and if it's having trouble attaching files or something, like you mentioned

brandly (approved):

PR #604, opened 2026-09-01
Regenerate types for the stream_idle_timeout generate-config field

### LINKED ISSUE #347: Timeout model calls when no streaming response received for a timeout period
https://inspectcommunity.slack.com/archives/C0805HJTK8U/p1787838999251509

Alex Remedios  [9:56 AM]
I'm finding that 16% of my model API calls are hanging against a 600s timeout. I have tuned this down to 300s (99th percentile of successful calls), but i wonder if this is an unaddressed case where we could detect streaming responses stopping and kill a streaming resopnse with no chunk for say 15s?

ransomr:
Regenerates the shared types for the new optional `stream_idle_timeout` field on `GenerateConfig`, and adds its entry to `effectiveConfig.ts`'s compile-time key record (the drift guard that keeps the unknown-field skip accurate).

Paired with UKGovernmentBEIS/inspect_ai#5096 ("Stalled streaming model calls now time out on chunk silence"), which adds the Python field. That PR's `check-schema-and-types` job is red until this merges.

Generated by `python src/inspect_ai/_view/schema.py` from the paired branch; `pnpm typecheck`, `pnpm test`, and `pnpm lint` all pass. The field is optional (Noneable), so the sibling scout app's duplicated types need no sync.

### DIFF
--- a/packages/inspect-common/src/types/generated.ts
+++ b/packages/inspect-common/src/types/generated.ts
@@ -2150,6 +2150,8 @@ export interface components {
             seed?: number | null;
             /** Stop Seqs */
             stop_seqs?: string[] | null;
+            /** Stream Idle Timeout */
+            stream_idle_timeout?: number | null;
             /** System Message */
             system_message?: string | null;
             /** Temperature */
--- a/packages/inspect-common/src/utils/effectiveConfig.ts
+++ b/packages/inspect-common/src/utils/effectiveConfig.ts
@@ -76,6 +76,7 @@ const GENERATE_CONFIG_KEYS: Record<keyof GenerateConfig, true> = {
   response_schema: true,
   seed: true,
   stop_seqs: true,
+  stream_idle_timeout: true,
   system_message: true,
   temperature: true,
   timeout: true,


dragonstyle (approved):
Correct generated-types regen for the new optional `stream_idle_timeout` field, and the matching `effectiveConfig.ts` key-record entry keeps the drift guard accurate. Both changes are mechanical and exactly what the schema regen produces. All checks green. Approving so the paired inspect_ai#5096 `check-schema-and-types` job goes green.

dragonstyle (approved):
Approve. Both lines are mechanical: the `generated.ts` hunk matches generator output for the new optional field, and the `effectiveConfig.ts` entry is the one addition the `Record<keyof GenerateConfig, true>` drift guard compiler-forces (typecheck is green, which proves it). Old logs are unaffected; the only behavior change is that live `stream_idle_timeout` retunes now show up in the viewer's effective config instead of being skipped as an unknown field. Land this before inspect_ai#5096, which then needs its submodule pin bumped to this PR's squash commit.

PR #608, opened 2026-09-02
Regenerate types for the logged_samples EvalResults field

### LINKED ISSUE #96: inspect ctl: task drain — stop dispatching new samples, let in-flight finish
Phase-3 directive from the control-channel design ([design/control-channel.md](https://github.com/meridianlabs-ai/inspect_ai/blob/main/design/control-channel.md) — "Phase 3 — modification (direct) methods" / "Other directives" and the endpoint table). One of the three remaining un-started phase-3 directives (alongside `sample requeue` and the per-sample limit knobs).

## Surface

- `POST /tasks/<task-id>/drain` — task-keyed like `config` / `log-flush` / `cancel`, so the handle never dangles across a retry (resolves the latest attempt via `latest_eval_for_task`).
- CLI: `inspect ctl task drain TASK`. Destructive verb, so `TASK` is required outright — no sole-running-task default (selector conventions in the design doc).
- Idempotent: draining an already-draining or finished task reports `changed: false`. Supports `--dry-run` and `--json` with the uniform mutation envelope `{target, applied, dry_run, detail}`.
- New route, so no `CONTROL_API_VERSION` bump: an older server answers with the stock `{"detail": "Not Found"}` 404 and the CLI verb passes `not_found_missing_route` to `_request_json` for the definitive "older inspect — restart the eval" message (same pattern as the cancel verbs).

## Semantics

- Stop dispatching queued samples; **in-flight samples finish naturally** — scored on their own terms, no interrupts. This is the complement of `task cancel --action score|error`, which *interrupts* in-flight samples; drain leaves them alone.
- Never-dispatched samples: presumably terminal `cancelled` in the counters and absent from the log, matching how cancel treats samples parked at the sample semaphore (checked as they leave the queue). Confirm in design.
- Eval-set interaction to settle: a drained task should presumably not be retried by the eval-set (treat it like the stamped `cancel_type` resolutions — a user cancel), but decide whether the final log status should let an explicit `eval-retry` re-run the undispatched samples.

## Two deferred items ride the same stop-dispatching machinery

The cancel-directives work deliberately deferred these to the drain implementation (see "Cancel a task / a sample" in the design doc):

1. **Graceful-drain cancel variant** — `task cancel` where in-flight samples finish naturally before the task ends (and a possible `--force` split between abort and graceful).
2. **Cancelling a pending eval-set retry** — `task cancel` on a task *between attempts* (`EvalState.retry_pending`) currently 409s with "re-issue the cancel once the retry is running"; actually cancelling the queued retry needs the same stop-dispatching hook.

Both should either land here or be explicitly re-scoped when this is designed.

ransomr:
Regenerates the shared types for the new optional `logged_samples` field on `EvalResults`.

Paired with UKGovernmentBEIS/inspect_ai#5093 ("inspect ctl: task drain — stop dispatching new samples, let in-flight finish"), which adds the Python field. That PR's `check-schema-and-types` job is red until this merges.

Generated by `python src/inspect_ai/_view/schema.py` from the paired branch (synced with inspect_ai `main`); `pnpm typecheck` and `pnpm test` (1234 tests) pass. The field is optional (Noneable), so the sibling scout app's duplicated types need no sync.

### DIFF
--- a/packages/inspect-common/src/types/generated.ts
+++ b/packages/inspect-common/src/types/generated.ts
@@ -1614,6 +1614,8 @@ export interface components {
             completed_samples: number;
             early_stopping?: components["schemas"]["EarlyStoppingSummary"] | null;
             headline?: components["schemas"]["HeadlineMetric"] | null;
+            /** Logged Samples */
+            logged_samples?: number | null;
             /** Metadata */
             metadata?: {
                 [key: string]: unknown;


dragonstyle (approved):
Verified this is the mechanical regen for `EvalResults.logged_samples` from inspect_ai#5093 — type, nullability, and position match the schema hunk there, and CI is green. Landing this first; #5093 will need its `_view/ts-mono` gitlink moved from `4e3ecca0` to the squash SHA once this merges.

PR #619, opened 2026-09-04
Fix main typecheck: add score_explanation to Scout transcript columns

brandly:
## Why main is red

The `typecheck` CI job has failed on every `main` commit since 3ccf7bf ("regenerate types"). That commit regenerated `apps/scout/src/types/generated.ts` and added an optional `score_explanation` field to `TranscriptInfo`, but `apps/scout/src/app/transcripts/columns.tsx` keeps three exhaustive `Record<keyof TranscriptInfo, ...>` maps (`COLUMN_LABELS`, `COLUMN_HEADER_TITLES`, `ALL_COLUMNS`), so `tsc` fails with TS2741 on each of them:

```
src/app/transcripts/columns.tsx(25,14): error TS2741: Property 'score_explanation' is missing in type '{ ... }' but required in type 'Record<"agent" | ... | "score_explanation" | ..., string>'.
```

## Fix

Register the field the same way every other transcript field is: a label, a header tooltip, a filterable string column (`value || "-"` cell, same shape as `error`/`limit`), and a slot in `DEFAULT_COLUMN_ORDER` right after `score`. It is deliberately **not** added to the default visible set, so the transcripts grid renders exactly as before until a user opts the column in via the column picker. No UI appearance change, so no screenshots.

The query builder in `src/query/transcriptColumns.ts` needs nothing: it is not exhaustive and its proxy already resolves unknown field names.

## Verification

- Reproduced the 3 TS2741 errors on the pre-fix file, then confirmed they are gone.
- `pnpm check` from the repo root: 33/33 tasks pass (manypkg, suppressions ledger, lint, typecheck, format).
- `pnpm --filter scout test`: 32 files, 280 tests pass.

No new suppressions.

https://claude.ai/code/session_01X5A9Hp44V9spvrx8MFjtiH

---
_Generated by [Claude Code](https://claude.ai/code/session_01X5A9Hp44V9spvrx8MFjtiH)_

### DIFF
--- a/apps/scout/src/app/transcripts/columns.tsx
+++ b/apps/scout/src/app/transcripts/columns.tsx
@@ -34,6 +34,7 @@ export const COLUMN_LABELS: Record<keyof TranscriptInfo, string> = {
   agent: "Agent",
   agent_args: "Agent Args",
   score: "Score",
+  score_explanation: "Score Explanation",
   metadata: "Metadata",
   source_id: "Source ID",
   source_type: "Source Type",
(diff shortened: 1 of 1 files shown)


brandly:
@review

PR #620, opened 2026-09-04
fix: validate element shapes at the eval-log normalization boundary

brandly:
The normalizers in `@tsmono/inspect-common/normalize` are the point after which raw eval-log JSON is treated as typed data and read unguarded (#555). They checked container types but not element shapes, so a log author could ship arrays and maps whose elements the generated types rule out and pydantic would refuse, and those elements reached readers written to the normalized contract:

- `model_fallbacks` with a null, number, or string element crashed `totalModelFallbacks` / `modelFallbackLines`. For a running eval under the local view server that threw while building the samples descriptor, replacing the whole viewer with the root error panel; for a settled log it threw inside `deriveSampleFields` during details ingestion.
- `results.scores` with a null or metrics-less entry, a sample `scores` map with a null value, or a message-list `input` with a non-message element threw in `deriveLogFields` / `deriveSampleFields` / `inputString` at ingestion.
- A full sample whose `input` array held a non-message element threw in `inputString` when the Scoring tab or the print view rendered.

The ingestion throws had a wider blast radius than the one file: `writeDetails` prepared every payload of a flush batch synchronously before any cache push or database write, so one malformed log kept every unrelated log in the same 250 ms flush window from reaching detailed depth, and the backfill re-enqueued them all together to fail the same way.

Changes, following the existing normalizer conventions (drop what pydantic would refuse, fill what it defaults, keep identity on clean input):

- A shared `normalizeModelFallbacks` drops elements without string `model` / `fallback_model`, fills a non-numeric `count` with 1, and turns a non-array into null. Used by both the summary and the full-sample normalizers.
- A shared `normalizeSampleInput` keeps string inputs, and for message lists drops entries whose `content` is neither a string nor a list of typed content items, dropping untyped items inside a message rather than the whole message.
- `normalizeSampleScores` drops score-map entries that aren't records with a `value`; a non-record map becomes null.
- `normalizeEvalResults` now requires each score to carry a string `name` (`scorer` backfills from it, as the v1 migration already does), fills `metrics` / `params` with `{}`, and drops metric entries without a numeric `value` (filling `name` from the map key and `params` with `{}`, since 2024-era logs wrote `options` instead).
- `writeDetails` prepares each payload of a batch on its own: a payload whose derivation still throws is logged and skipped while the rest of the batch lands.

No read-site guards were added; the readers keep trusting the normalized types. No new lint suppressions: the new checks are type predicates, and the per-file directive counts are unchanged.

Tests: table-driven cases in `summary.test.ts` and `normalize.test.ts` feed each malformed shape through the normalizer and then through the reader that used to throw; `logsContent.test.ts` writes a two-log batch where one log's derivation throws and asserts the other still reaches the database at detailed depth. Each test failed against the previous code.

Not in this PR: the skipped log is still re-enqueued by the next backfill (it is never recorded as a failed fetch, since attempt bookkeeping lives in the fetch engine), and a throw from `toLogPreview` inside `onDetailsComplete` can still end a queue worker. Both need engine-side changes and are left as follow-ups.

### DIFF
--- a/apps/inspect/src/client/utils/normalize.test.ts
+++ b/apps/inspect/src/client/utils/normalize.test.ts
@@ -49,6 +49,24 @@ describe("normalizeEvalHeader", () => {
     expect(header.metadata).toEqual({ from: "log" });
   });
 
+  it("fills stats usage defaults and drops malformed usage entries", () => {
+    const header = normalizeEvalHeader({
+      eval: minimalEval,
+      stats: {
+        model_usage: {
+          "openai/gpt-4": { input_tokens: 1, output_tokens: 2 },
+          "openai/gpt-3.5": null,
+        },
+      },
+    });
+    expect(header.stats?.model_usage).toEqual({
+      "openai/gpt-4": { input_tokens: 1, output_tokens: 2, total_tokens: 0 },
+    });
+    expect(header.stats?.role_usage).toEqual({});
+    expect(header.stats?.started_at).toBe("");
+    expect(normalizeEvalHeader({ eval: minimalEval }).stats).toBeUndefined();
+  });
+
   it("preserves fields it doesn't model (future schema growth)", () => {
     const header = normalizeEvalHeader({
       eval: minimalEval,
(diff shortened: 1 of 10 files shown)


brandly:
@review

PR #621, opened 2026-09-04
fix: stop log-authored keys from reaching Object.prototype in the viewer

brandly:
## Summary

Several places in the inspect viewer (and one in Scout) used strings written by the log author — event uuids, model and role names, score and scorer names, scanner reference ids, skeleton span parents, attachment ids — as keys on plain objects or arrays. A name such as `__proto__`, `constructor` or `toString` then either resolved through the prototype chain to a builtin, or worse, wrote through it. Two of these were session-wide prototype pollution; the rest were deterministic render crashes to the root error panel.

Session-wide pollution:

- **Chunked skeleton index** — a span whose `parent` was `"__proto__"` indexed `childrenOf` through `Array.prototype` and pushed onto it, changing every past-the-end array read in the viewer (chunk bounds, cursor buffers, outline rows) for the rest of the session, including for other logs opened afterwards. `SkeletonIndex` now requires `parent` to be an integer index of an earlier span and throws otherwise; `at` accepts only own integer positions.
- **Usage config builders** — a model or role named `__proto__` wrote every key of the log's `model_args` / generate config onto `Object.prototype`; one named `constructor` overwrote statics on `Object` (`model_args: {"keys": ...}` broke `Object.keys`). Reached from the Models tab, every sample's Usage tab and the Timeline tab. The builders now accumulate in Maps and return null-prototype records.
- **Property bags (inspect app slice and Scout store)** — transcript event panels store per-panel UI state in a bag named by the event uuid. Creating a bag named `__proto__` inside the immer recipe assigned the tab selection onto the real `Object.prototype`, and because reads used `key in bag`, every other event panel then inherited it and rendered empty until reload. Both stores now read bags and keys as own properties only and create a missing bag with a computed-key literal (plain assignment of a `__proto__` name goes to the inherited setter, which immer turns into a `setPrototypeOf` error). The bags stay plain records because the slice is persisted as JSON.

Inherited-key reads that crashed the viewer or substituted a builtin:

- **Info tab** — a scorer named after an `Object.prototype` member made `PlanDetailView` call `.push` on the inherited member. Grouping is now a Map.
- **Samples grids** — a score named `constructor` made the column builder call `.add` on the `Object` function while discovering cross-log columns, so one such log broke the directory's Samples view; label and colour-scale overrides were read the same way. Tallies are Maps; the override maps are read as own properties.
- **Scan reference previews** — a scanner reference id naming a prototype member yielded a builtin as `citePreview`, and hovering the cite called it during render. The preview table is a Map.
- **Transcript event labels** — an event uuid naming a prototype member got the builtin as its label whenever a search result supplied any event citation, and `label.replace` threw. Event, message and tool label maps are read as own properties.
- **attachment:// resolution** — `attachment://constructor` resolved to the `Object` function and was substituted into message content, which the chat renderer then tried to iterate. Attachment lookups are own-key only, so these are ordinary misses.

## Approach

A small `getOwn` helper (own-property read of a string-keyed record) and `nullProtoRecord` (moved from `connectionHistory.ts`, which already did this) live in `@tsmono/util`; keyed accumulation uses `Map`. Each change has a test that fails on the previous source: tests plant keys on the real `Object.prototype` / `Array.prototype` / `Object` inside `try`/`finally` or `afterEach`, assert that nothing leaks, and assert that lookups of `constructor` / `toString` / `__proto__` read as absent rather than as a builtin.

Two `no-unnecessary-condition` suppressions in `configsForUsage.ts` became unnecessary and were removed (`suppressions.json` shrinks by two entries).

## Verification

`pnpm check`, `pnpm test`, and `pnpm build` all pass from the repo root.

### DIFF
--- a/apps/inspect/src/app/plan/PlanDetailView.test.tsx
+++ b/apps/inspect/src/app/plan/PlanDetailView.test.tsx
@@ -1,0 +1,53 @@
+import { cleanup, render, screen } from "@testing-library/react";
+import { afterEach, describe, expect, it } from "vitest";
+
+import { testEvalScore, testEvalSpec } from "@tsmono/inspect-common/testing";
+import type { EvalScore } from "@tsmono/inspect-common/types";
+import {
+  ComponentIconProvider,
+  ComponentNavigationProvider,
+} from "@tsmono/react/components";
+import { ComponentStateProvider } from "@tsmono/react/state";
+import { makeStateHooks, testIcons } from "@tsmono/react/testing";
+
+import { PlanDetailView } from "./PlanDetailView";
+
+afterEach(cleanup);
+
+const renderScores = (scores: EvalScore[]) =>
+  render(
+    <ComponentStateProvider hooks={makeStateHooks()}>
+      <ComponentIconProvider icons={testIcons}>
+        <ComponentNavigationProvider navigation={{ navigate: () => {} }}>
+          <PlanDetailView evaluation={testEvalSpec()} scores={scores} />
+        </ComponentNavigationProvider>
+      </ComponentIconProvider>
+    </ComponentStateProvider>
+  );
+
+describe("PlanDetailView scorers", () => {
+  it("groups scores by scorer", () => {
+    renderScores([
+      testEvalScore({ scorer: "match", name: "accuracy" }),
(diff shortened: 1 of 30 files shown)


brandly:
@review

brandly:
@review

